PR #60, opened 2019-11-16
Readme updates and stop regression auto updates 

alainmarcel:


hzeller:
FYI, if you want to test README markdown, you can use the grip tool https://github.com/joeyespo/grip to render it locally, so it is easy to see if it is rendering properly.

PR #73, opened 2019-11-18
Staging for make install

alainmarcel:
Replaced release.tcl with cmake


hzeller:
Travis is not happy:
```
 while executing
"exec sh -c "cd $REGRESSION_PATH/tests/$test/; rm -rf slpp*""
    (procedure "run_regression" line 62)
    invoked from within
"run_regression"
    invoked from within
"set result [run_regression]"
    (file "../tests/regression.tcl" line 577)
```

PR #75, opened 2019-11-18
Link tcmalloc (looks like that was lost converting to cmake)

hzeller:


alainmarcel:
That would explain the sudden slow down, not the latest flatbuffer! Good catch!

PR #82, opened 2019-11-19
Make local includes relative to base

hzeller:


hzeller:
Through a typing glitch, the first commit co-incides with #81, but if that is merged first, this should not create an issue.

Issue #89, opened 2019-11-20
Slow parsing of scr1


hzeller:
As already discussed offline, but recorded here, so that we can attach commits to it.

To reproduce:
```
git clone https://github.com/syntacore/scr1.git
surelog -nobuiltin -parse -noelab -Iscr1/src/includes scr1/src/pipeline/scr1_pipe_hdu.sv scr1/src/pipeline/scr1_pipe_tdu.sv scr1/src/pipeline/scr1_ipic.sv scr1/src/pipeline/scr1_pipe_csr.sv scr1/src/pipeline/scr1_pipe_exu.sv scr1/src/pipeline/scr1_pipe_ialu.sv scr1/src/pipeline/scr1_pipe_idu.sv scr1/src/pipeline/scr1_pipe_ifu.sv scr1/src/pipeline/scr1_pipe_lsu.sv scr1/src/pipeline/scr1_pipe_mprf.sv scr1/src/pipeline/scr1_pipe_top.sv scr1/src/core/primitives/scr1_reset_cells.sv scr1/src/core/primitives/scr1_cg.sv scr1/src/core/scr1_clk_ctrl.sv scr1/src/core/scr1_tapc_shift_reg.sv scr1/src/core/scr1_tapc.sv scr1/src/core/scr1_tapc_synchronizer.sv scr1/src/core/scr1_core_top.sv scr1/src/core/scr1_dm.sv scr1/src/core/scr1_dmi.sv scr1/src/core/scr1_scu.sv scr1/src/top/scr1_dmem_router.sv scr1/src/top/scr1_imem_router.sv scr1/src/top/scr1_dp_memory.sv scr1/src/top/scr1_tcm.sv scr1/src/top/scr1_timer.sv scr1/src/top/scr1_mem_axi.sv scr1/src/top/scr1_top_axi.sv scr1/src/pipeline/scr1_tracelog.sv scr1/src/tb/scr1_memory_tb_axi.sv scr1/src/tb/scr1_top_tb_axi.sv
```

alainmarcel:
Fixed by #103

PR #103, opened 2019-11-22
Fix scr1 preproc runtime, simplify pp grammar

alainmarcel:
Signed-off-by: Alain <alainmarcel@yahoo.com>


alainmarcel:
Fixes #89

PR #184, opened 2020-02-17
Don't redeclare ANSI ports in OneNet tests

rkapuscik:
IEEE Std 1800-2017 23.2.2.2: The port identifier shall not be
redeclared, in part or in full, inside the module body.

Signed-off-by: Rafal Kapuscik <rkapuscik@antmicro.com>


alainmarcel:
We need to support this old declaration style too.
All commercial tools still accept it with a warning. I'll rename the test OneNetNonAnsi,
And create the Ansi version: OneNet

PR #194, opened 2020-02-28
Find tcmalloc library using the cmake find_library() feature.

hzeller:
This hopefully makes it more reliably compile in some
environments (e.g. currently in building the conda packages we
have to switch off the tcmalloc feature because it doesn't build
there).

Also ran cmake-format, which formatted two or three lines.


hzeller:
Running conda locally, this indeed seems to fix the conda package issue. So this means we can use tcmalloc again and be faster.

PR #332, opened 2020-04-29
Make dependency on capnp and libkj self-contained.

hzeller:
Always make sure to link the Capnproto version that was
compiled with UHDM to avoid compatibility issues by
accidentally linking it to a version that might be on the
host system.

While at it: update to latest UHDM that has tighter
control on installing these libraries.


hzeller:
Ah, this requires the newer cmake version now. Let me fix that first.

hzeller:
In #333 this will happen.

hzeller:
Alright, looks like we're ready to merge.

Issue #331, opened 2020-04-29
Unable to provide own Python listener script


mkurc-ant:
While trying to write an own python listener I've encountered actually two issues:

1. The `-pythonlistenerfile` commandline argument is parsed but its value is never actually used in the code. Besides that IMO there is a bug: there should be `i+1` instead of `i`: https://github.com/alainmarcel/Surelog/blob/19c61c6377060b0354b8f39d440c2ec7ec714cdc/src/CommandLine/CommandLineParser.cpp#L841

2. This is the part of code that loads the listener script: https://github.com/alainmarcel/Surelog/blob/19c61c6377060b0354b8f39d440c2ec7ec714cdc/src/API/PythonAPI.cpp#L178-L200

    I don't quite understand the intention: It looks like the code is supposed to look for the script with the fixed name and try to load it. If it fails then it moves to a next fixed file name and tries that etc. But it actually works differently by loading all the scripts that it finds and setting `m_listenerScript` to the last one found.

    I've tried to supply my own `slSV3_1aPythonListener.py` file with some callback functions defined. Either my file wasn't loaded / parsed correctly or it's a `PyRun_SimpleFile()` thing which does not override already defined functions (?) In the end my functions weren't called at all.

By hacking the code and making it load explicitly the file I wanted it all worked correctly.

I can fix both issues in the code (and make a PR) but first I'd like to know if I'm correct about the bugs / shortcomings.

alaindargelas:
Please do submit all the changes you mentioned.
The absolute paths to some installation are obsolete and should be removed. Surelog does not install this file anymore. The command line argument was probably not tested at all, please fix it.
The only thing that can still make sense to keep is the local path to the directory.
By the command line should take precedence. 
I think there is one test Surelog/tests/ that uses the path in the local directory.
Thanks!

PR #337, opened 2020-04-30
Python listener loading fixes

mkurc-ant:
This PR fixes some issues with providing custom Python listener. Fixes https://github.com/alainmarcel/Surelog/issues/331

The problem with the command line option `-pythonlistenerfile` was that its value was used after all python scripts were loaded. So I modified script loading in Python API so that they are loaded after command line parsing.

I also removed obsolete fixed paths to python scripts. By default Python scripts are loaded from `<program path>/python` unless a script with the same name is present in the current directory. Apart from that the listener script is loaded first from the file specified in the command line option.


alaindargelas:
Thanks!
BTW, if you are interested by the teask, we need a Python API to wrap the UHDM API(s), maybe the VPI (C) API or the C++, or both.

mkurc-ant:
@alaindargelas Thanks, a Python API for UHDM looks interesting but I don't spare have to contribute to that now.

Issue #239, opened 2020-03-28
Windows/Mac Support?


hs-apotell:
Any plans to support Windows and Mac platforms?

alainmarcel:
No plans, but volunteers to add support are welcome!
We have to check the support from the following components used by Surelog first:
antlr
flattbuffers
capnp-proto
google-tests

hs-apotell:
All of the above mentioned libraries successfully build on Windows. Running into Surelog specific issues now.

src/API/SLAPI.h - There are two files with the same name, if casing is ignored. Windows is case-insensitive and so one file has to either be renamed or has to move into a subfolder.

SLAPI.h is generic api i.e. non-language specific. Can this be renamed to api.h?

src/API/slapi.h is auto-generated by src/API/embed_python_api.tcl and is very python specific. Can we call this file slpyapi.h or move into subfolder src/API/py/slapi.h? May be even move all the python related files in this folder into the subfolder?

Any other suggestion as to what's an acceptable resolution to move forward?

hs-apotell:
Is there a specific version of C++ Surelog supports? Is C++17 and above acceptable?

Reasoning - Some of the posix functions don't have equivalent functions under VC. However, C++17 added some of the functionality that can be used to replace and will work across platforms. For instance, use of file io functionality in src/Utils/FileUtils.cpp.

alaindargelas:
Windows support achieved, Thanks @hs-apotell!
Do we need a travis Windows build?

PR #372, opened 2020-05-13
Issue #239: Add support for Windows

hs-apotell:
Issue #239
Add support for Windows

Highlights -
Renaming slapi.* files to slapi_scripts.* to support non-case
sensitive platforms like windows. The folder already contains
another file with the same name and that won't scale to these
platforms. Files renamed and related scripts updated.


hs-apotell:
What was the issue with the Presubmit checks failing? Could you provide some insight please!

alaindargelas:
One of the tests (Ariane) does crash sometimes when ran in parallel with other tests in the presubmit (Google kokoro system). It runs fine one test at a a time in Travis.

  On Wed, May 13, 2020 at 16:13, HS<notifications@github.com> wrote:   


What was the issue with the Presubmit checks failing? Could you provide some insight please!

Issue #376, opened 2020-05-14
uhdm-dump not anymore where regression.tcl expects it (with fresh UHDM)


hzeller:
The regression.tcl expects that the uhdm-dump binary is at the same place installed where surelog is in the dist ( https://github.com/alainmarcel/Surelog/blob/master/tests/regression.tcl#L155 ).
(That would be `build/dist/Release/uhdm-dump`).

However, with latest UHDM, it is not there anymore, but only to be found in `./build/third_party/UHDM/uhdm-dump`

As a result, we see FAILDUMP results:
https://travis-ci.com/github/alainmarcel/Surelog/builds/165906929#L1899

Not sure what is different, so not clear about a good solution yet: maybe the more restrictive CMake change in https://github.com/alainmarcel/UHDM/pull/228 prevents it from being there ? Should we just point the `UHDM_DUMP_COMMAND` to the location in third_party/UHDM ?

alaindargelas:
@hs-apotell the changelist alainmarcel/UHDM#228 broke Surelog regression. The last stage of the regression is to do a uhdm-dump, the executable has changed place with your changelist.
I don't understand the rational for changing the location of the tests executable in UHDM. We need to reconcile what you are trying to achieve on Windows and how the overall system needs to be setup: UHDM and Surelog with UHDM as a subsystem.
I will override your changes until you give an explanation on why the previous setup did not work on Windows.

hs-apotell:
Sorry about the regression. I was supposed to follow up with accompanying surelog change but got sidetracked with the crash on UHDM serialization (Windows only). I have an open pull request to bring both projects back in sync and also fix the crash.

The rationale behind the change was that Visual Studio and QT Creator, both doesn't play well when the output files are moved around. They expect these files to be in specific location. Plus, on Windows, CMake works in multi-configuration mode which brings in its own challenges. The whole change started with me fixing the following scenario -
* Target surelog (library project) generates "surelog.lib" in the root output folder
* Target surelog_bin (binary/executable project) also generates "surelog.lib" in the root output folder.

The problem is that one of the library that surelog executable links to is exporting symbols which the linker exports as the library (unusable output but the linker still does it because it doesn't differentiate between a dll and executable - required the former but useless for the latter). If you haven't heard of DLL hell on windows, you should read up on it. It really is what the name implies.

[...]

PR #377, opened 2020-05-14
Use uhdm-dump from UHDM compile.

hzeller:
This is an attempt to work-around #376

Not sure if this is the right solution, so mostly for discussion.


hzeller:
FYI, To use with a fresh UHDM, we also need https://github.com/alainmarcel/UHDM/pull/234 because once the script finds the `uhdm-dump`, we run into the trouble of Tcl catch/exec failing if there is some text to stderr...

alaindargelas:
Not needed

PR #379, opened 2020-05-15
Issue #376: Fix to incorporate changes to UHDM cmake file.

hs-apotell:
Issue #376
Fix to incorporate changes to UHDM cmake file.

Highlights -
* Generate output files in location where the dev tools expect
  instead of moving them to custom location. Rationale being that
  dev tools work better when the files aren't moved around.
  Visual Studio fails to find the generated test binaries if
  they aren't in the expected location. Plus, on windows, CMake
  works in multi-configuration mode and so having a single "dist"
  folder doesn't work well.
* Upgrade minimum CMake version required to 3.15
  This is required for the policy CMP0091 which was introduced in
  this version.
* Also, upgraded minimum C++ version to 17 to support filesystem
  module. This change will help cross-platform support.

IMPORTANT: This change is blocked by -
https://github.com/alainmarcel/UHDM/pull/236

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 # -*- mode:cmake -*-
-cmake_minimum_required(VERSION 3.15)
+cmake_minimum_required(VERSION 3.5.1)
 
 # Detect build type, fallback to release and throw a warning if use didn't
 # specify any
@@ -30,7 +30,7 @@ add_subdirectory(third_party/antlr4/runtime/Cpp EXCLUDE_FROM_ALL)
 add_subdirectory(third_party/UHDM)
 add_subdirectory(third_party/googletest EXCLUDE_FROM_ALL)
 
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 11)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,13 +1,5 @@
 # -*- mode:cmake -*-
-cmake_minimum_required(VERSION 3.5.1)
-
-set(FLATBUFFERS_BUILD_TESTS OFF CACHE BOOL "Skip flatbuffers' tests")
-add_subdirectory(third_party/flatbuffers EXCLUDE_FROM_ALL)
-add_subdirectory(third_party/antlr4/runtime/Cpp EXCLUDE_FROM_ALL)
-add_subdirectory(third_party/UHDM)
-add_subdirectory(third_party/googletest EXCLUDE_FROM_ALL)
-
-find_package(SWIG REQUIRED)
+cmake_minimum_required(VERSION 3.15)
 
 # Detect build type, fallback to release and throw a warning if use didn't
 # specify any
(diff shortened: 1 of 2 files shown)


hs-apotell:
NOTE: This merge has to be accompanied with upgrade to UDHM pull request https://github.com/alainmarcel/UHDM/pull/236

alaindargelas:
I cannot accept this changelist in its current state. Too many moving parts.
Please remove the changes about cmake and gcc version. There are many clients systems that we need to check for readiness for this upgrade in compiler and cmake. This has the potential to disrupt many users, myself included, I don't have time to upgrade my systems for next week at least.
In general system-level changes should come in piecemeal, not in bulk.
Thanks

hs-apotell:
CMake & C++ upgrade change pulled back. Will reintroduce the change in a later commit with working solution for Windows platform.

hzeller:
Thanks, I think overall it looks like it moves in the right direction.

I agree with Alain though, that small changes will help to make it easier to review and test (the cmake implications across multiple sub-projects to be compiled can sometimes be surprising; I am a friend of good ol' Makefiles, so I have not fully grokked yet all the subtle ways cmake interacts with multiple projects).

With regard to version: commonly you find cmake 3.13 on Debian stable, so if we can we should first address all things that can be done with that version to have a good out-of-the-box experience (similar in UHDM).
Are there ways were we can request a minimum version of 3.13, but then, if a newer version is available, select the features we then can use ?
Not a total showstopper to use 3.15, but it would be much easier to get people involved if they can use build-tools right there; so if we can select needed MSVC features if that newer version of cmake is installed there, the better.

But first things first: do you think the change can be broken down into smaller atomic parts ?
Thanks

hs-apotell:
> ... that small changes will help to make it easier ...

I agree smaller the better. The changes in this commit are mostly tied to one over arching goal - support for transitive dependencies. I don't see how I will implement it for one dependency and not other. That would make it harder to test. 

> ... a minimum version of 3.13 ...

3.15 is minimum required to support runtime library and this policy implements it [CMP0091](https://cmake.org/cmake/help/latest/policy/CMP0091.html). The issue here is that some of the dependent libraries do not implement /MT option (they use /MD which is the default). The alternative solution would be to modify these dependencies. Use of submodules (and not have a copy of it in the repo) makes that harder.

I have pulled back the CMake and C++ version upgrade since the windows as a platform isn't supported officially until it works.

alaindargelas:
Race condition?All of these tests are ran in parallel

  On Fri, May 15, 2020 at 12:27, Henner Zeller<notifications@github.com> wrote:   


mmh, so the regression test seems to no mysteriously fail with not finding the surelog binary (look at the cmake output towards the end before the regression table; where it generates the regression test targets) https://source.cloud.google.com/results/invocations/e0772462-337f-4bc9-8780-fb1cd66a6286/targets/foss-fpga-tools%2Falainmarcel-surelog%2Fpresubmit/log
ERROR: Cannot find executable /tmpfs/src/github/alainmarcel-surelog/build/dist/Release/surelog!
ERROR: Cannot find executable /tmpfs/src/github/alainmarcel-surelog/build/dist/Release/surelog!
ERROR: Cannot find executable /tmpfs/src/github/alainmarcel-surelog/build/dist/Release/surelog!

However, the final table is still produced, and possibly the issues are due to segfaults, and the surelog binary thing is just a red herring ?
Something definitely weird here.

alaindargelas:
The message is outputted by individual tests checking for the executable existence at the beginning of regression.tcl, now regression.tcl is executed by cmake in parallel.It only fails for some tests, maybe the first to execute and the exec is still not visible, file system issue?

  On Fri, May 15, 2020 at 12:27, Henner Zeller<notifications@github.com> wrote:   


mmh, so the regression test seems to no mysteriously fail with not finding the surelog binary (look at the cmake output towards the end before the regression table; where it generates the regression test targets) https://source.cloud.google.com/results/invocations/e0772462-337f-4bc9-8780-fb1cd66a6286/targets/foss-fpga-tools%2Falainmarcel-surelog%2Fpresubmit/log
ERROR: Cannot find executable /tmpfs/src/github/alainmarcel-surelog/build/dist/Release/surelog!
ERROR: Cannot find executable /tmpfs/src/github/alainmarcel-surelog/build/dist/Release/surelog!
ERROR: Cannot find executable /tmpfs/src/github/alainmarcel-surelog/build/dist/Release/surelog!

However, the final table is still produced, and possibly the issues are due to segfaults, and the surelog binary thing is just a red herring ?
Something definitely weird here.

hzeller:
But I wonder where the reference to `build/dist/Release/surelog` is coming from it complains about while it is in `build/bin/surelog`. anyway, investigating.

hzeller:
Fixed in #383

PR #383, opened 2020-05-15
Surelog is now generated in build/bin. Update path to where regressio…

hzeller:
…n test expects it.


hs-apotell:
@hzeller Thanks for spotting that.

PR #384, opened 2020-05-15
Bump CMake required version to 3.13 to be consistent with UHDM requir…

hzeller:
…ements.


hzeller:
(this doesn't Travis yet, because #383 that fixes the build is not merged yet)

PR #401, opened 2020-05-21
Issue #239 (Part 1/4) - Add Windows platform support

hs-apotell:
Issue #239 (Part 1)
Add Windows platform support

Highlights -
* Raise the minimum CXX standard to 14 (this matches what UHDM's requirement)
* Use Python3 package (instead of PythonLibs) to get the runtime library directory
  path. This is required to locate the specific python dll needed to deploy for
  DLL specific platforms like windows.
* For windows specific builds, move the generated library for the surelog binary
  target to the bin folder to avoid collision with the surelog library target.
* For unix only, add dependency on stdc++fs to bring in support for filesystem
  module with gcc7.
* Split the post-build step into two - platform dependent and platform independent
  This is required because Windows doesn't have support for symlinks and so the
  only viable option is to copy over the necessary files. Common post-build would
  penalize unix platform users.
* When installing, use target outputs without naming them so the commands work
  across platforms.
* Decorative change - Fixed a few indenting issues


CMakeLists.txt
```diff
@@ -346,23 +354,47 @@ target_link_libraries(hellouhdm surelog)
...
+            ${Python3_RUNTIME_LIBRARY_DIRS}/python3${Python3_VERSION_MINOR}$<$<CONFIG:Debug>:_d>.dll
+            ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}
+    COMMAND echo "       Platform depdendent staging completed"
+    WORKING_DIRECTORY "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}")
+else()
+  add_custom_command(
+    TARGET surelog-bin
+    POST_BUILD
+    COMMAND echo "       Creating platform depdendent staging for precompiled packages"
+    COMMAND ln -fs ${PROJECT_SOURCE_DIR}/third_party/UVM/ovm-2.1.2 ovm-2.1.2
```
hzeller, line 391: Question for @alaindargelas : for simplicity and ease to maintain different platforms, should we maybe just have always the copy_directory version from above ? It means a little wasted disk-space and time for Unix builds, but I think the unifomrity and maintainability is worth it.

CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ add_subdirectory(third_party/antlr4/runtime/Cpp EXCLUDE_FROM_ALL)
 add_subdirectory(third_party/UHDM)
 add_subdirectory(third_party/googletest EXCLUDE_FROM_ALL)
 
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 14)
```
hzeller, line 33: Do we need this for technical reasons or would only change it because UHDM also requires it ? (it might make sense to also look if we actually need C++14 there).

14 was an intermediate standard with not much changed since 11; 11 is desirable as many folks working in the hardware-related field often have super-old cent-os systems and sometimes struggle to get a newer compiler, so staying a few generations behind in the requirements opens it for more people.

So I am happy to accept a C++14 change if this is really needed to write more manageable code.

CMakeLists.txt
hs-apotell: Ideally, we would do neither copy nor link. We should be able to provide all these paths as inputs and the binary should pick it up. Is there a reason why all of these have to be under one single root folder? Why can't we use the UVM & OVM from in place in third party folder? Sane goes for the scripts as well. Is there a reason why we can't just list them as command line arguments to the tool.

hzeller: Yes good question - @alaindargelas is the right person to answer that.

CMakeLists.txt
hzeller: So, looks like capnproto actually wants C++14, so in turn that means UHDM needs it. With that, I guess we can leave this here as minimum requirement.

hs-apotell: Ideally, I would go forward rather than backwards i.e. move to 17 instead of 11. As you said, v14 was an intermediate and things weren't locked down with the compilers. v17 is more polished. v11 lacks the platform support needed to keep the code clean of #if/#endif. Multiple cross platform features including filesystem, threads, and atomic moved out from non-experimental phase with this upgrade. Using experimental features is not ideal, especially for production code. But, that was the highest/easiest I could get to without raising eyebrows (because UHDM is already at 14) and still get some of the features needed to support other platforms.

> ... have super-old cent-os systems ...

I wouldn't disagree with that. I have had experienced this first hand. Given what Surelog/UHDM is (and wants to be), I contend that we have already made our choices.

* Using mostly C++ instead of primitive C
* Use of ANTLR for parsing (which happens to be Java based) instead of using flex/bison
* Use of python v3 instead of v2
* Choice of gcc7

Those choices speak for themselves. Those older systems you are talking about are already too old for the choices we have made. We have chosen what's current/active instead of what's the industry norm. I am not hearing an argument for holding back now.

I would also like to point out that Surelog is dependent on other third party modules (flatbuffers, googletest, and few others). Our hands might be tied if any one of those external dependencies decided to stop support for older standards or compilers. If anything we should be ahead of them not lagging in embracing newer technologies/standards. I surely do not wish to spend time reimplementing cross-platform filesystem IO or thread module.

Our focus should be on "completeness" of the parser and the technology, in general. If it can support all the old (and very old, in some cases) VHDL/Verilog code, that would be a bigger win then focusing our attention on which specific compiler or C++ standard to use. Nobody wants to reimplement what was implemented 30 years back. That stuff should just work because it has worked for three decades. You touch it and you break it.

All that said, if we decide to stick to v11, few class implementations would need to be changed. Surely, not the end of the world. A few coding guidelines to avoid sprinkling platform dependent code all over the source code would suffice.

> So I am happy to accept a C++14 change if this is really needed to write more manageable code.

This will be minimum required for filesystem support i.e. experimental filesystem support. Ideally, move up to C++17 to get non-experimental/production quality support from compilers.

@alaindargelas, @hzeller That's my penny! Will leave it to other stakeholders for final call.

hzeller: So personally, I'd go all the way to 17 as well for the std::filesystem reasons. And if there is a good time, it is now :)
I'll leave it to Alain to decide.

hs-apotell:
@alaindargelas @hzeller The CI test failure - Is this one of those failing because of running parallel build or is this a genuine failure? This commit only has changes to CMakefile.txt and the error in the log looks like it failed for some parsing reason. I fail to connect the dots. Appreciate some direction on this.

hs-apotell:
https://github.com/alainmarcel/Surelog/pull/402 includes the change in this pull request. All checks succeeded for the latter. I assume the failure on this pull request is volatile.

Please comment if this needs attention.

CMakeLists.txt
alaindargelas: Because we don't want any relative paths of any sort saved in the database. If these files are not under the current dir, you'll have some form of relative path. This is for precompiled packages that should be "portable", meaning, no relative paths. You'd have to perform some hacking to mask the relative path of the files.

alaindargelas:
The only volatile failure is the Ariane testcase.
All the other failures (3 tests failing + one executable not being linked correctly) are real issues that need to be fixed prior to merging.
- The testscases behavior change are due to new code not working correctly, I suspect some file utility function you changed.
- The test executable not linking is most likely due to you forgetting to update that link line in the top level Makefile.

You should see all of these issues in your local build when you run  the following:
rm -rf build
make release -j
make regression

CMakeLists.txt
alaindargelas: The copy systematically does not work for me, it would have to copy only if the files are not there already, else the time stamp would change, triggering a recompilation of the pre-compiled  packages (which is what this step is performing). That is an expensive task that should only be performed when the grammar files change in one's build enviroment.

alaindargelas:
> [ FATAL] : 02724> [ SYNTAX] : 42725> [ ERROR] : 422726> [WARNING] : 182727> [ NOTE] : 5327285,7c356,3582729< |SYNTAX | 6 | 6 |2730< | ERROR | 24 | 26 |2731< |WARNING| 19 | 11 |2732---2733> |SYNTAX | 6 | 4 |2734> | ERROR | 24 | 42 |2735> |WARNING| 19 | 18 |27369c3602737< | NOTE | 12 | 26 |2738---2739> | NOTE | 12 | 53 |
Above it shows that you broke the functionality of the "diff mode" done with the fork (main.cpp change)

Below, a file is not discovered (not processed), hence the testcase behaves dramatically differently (I suspect you broke some utility functions):


============================== DIFF ======================================================1950diff /home/travis/build/alainmarcel/Surelog/tests/UnitLibrary/UnitLibrary.log tests/UnitLibrary/UnitLibrary.log195111,12d101952< [INF:PP0122] Preprocessing source file "/home/alain/Surelog/tests/UnitLibrary/libwconfig/libw.cfg".1953< 

Same for :| LibraryIntercon | DIFF | 0 | 0 | 0 | 0 (-3) | 6 (-24) | 0s | 27 |




    On Thursday, May 21, 2020, 02:22:13 PM PDT, HS <notifications@github.com> wrote:  
 
 


#402 includes the change in this pull request. All checks succeeded for the latter. I assume the failure on this pull request is volatile.

Please comment if this needs attention.

CMakeLists.txt
hs-apotell: Yes that is going to be a problem at the moment since copy_directory command always overwrite the destination contents regardless. Here's a custom implementation - an attempt to implement update_directory command - which can be customized to our needs.

https://bravenewmethod.com/2017/06/18/update_directory-command-for-cmake/

Possibly an exercise for some other day!

alaindargelas:
Yes, that is fine, we can move to 17. But the current implementation in the changelist has one or more bugs that change the behavior of 3 tests.So, I suggest to undo the C++ changes, submit the Cmake changes, then retry the C++ changes with the filesystem.

    On Thursday, May 21, 2020, 12:58:44 PM PDT, Henner Zeller <notifications@github.com> wrote:  
 
 


@hzeller commented on this pull request.

In CMakeLists.txt:
> @@ -30,7 +30,7 @@ add_subdirectory(third_party/antlr4/runtime/Cpp EXCLUDE_FROM_ALL)
 add_subdirectory(third_party/UHDM)
 add_subdirectory(third_party/googletest EXCLUDE_FROM_ALL)
 
-set(CMAKE_CXX_STANDARD 11)
+set(CMAKE_CXX_STANDARD 14)

So personally, I'd go all the way to 17 as well for the std::filesystem reasons. And if there is a good time, it is now :)
I'll leave it to Alain to decide.

Issue #431, opened 2020-05-27
Kokoro build breaks because of Python


alaindargelas:
Started breaking after around  #239, one of the Cmake changes changed Python and Kokoro is not happy with it.

alaindargelas:
Fixed, thanks @hs-apotell!

PR #432, opened 2020-05-27
Issue #431: Force the pyenv to use specific version

hs-apotell:
Issue #431 
Force the pyenv to use specific version


hs-apotell:
@alaindargelas Could you please kick off another verification build to confirm the change fixes the issue. The last build failed for some off PGP issue and I can't seem to find a way to kick off builds without submitting another change.

PR #475, opened 2020-06-04
Minor cleanup of README

mgielda:
Signed-off-by: Michael Gielda <mgielda@antmicro.com>


mgielda:
Tried to also change section header levels to a meaningful structure (group usage-related sections together, keep only title on level 1, make last section level 2 - since it's not usage related)

Issue #410, opened 2020-05-23
FileUtils::collectFiles is not thread safe


hs-apotell:
Use of `chdir` in implementation is likely to cause non-reproducible errors, especially in multi-threaded environment.

Suggestion - FileUtils::collectFiles(std::string, SymbolTable) is making assumptions about what the base/starting path should be (and also changing to root folder when the input path is absolute). Why not take that the base, non-wildcard path as the argument?

alaindargelas:
Please propose another implementation. It has to match the spec page 904 of the 2017 SystemVerilog Standard:

Library map file details:
1) file_path_spec uses file-system-specific notation to specify an absolute or relative path to a
particular file or set of files. The following shortcuts/wildcards can be used:
? single character wildcard (matches any single character)
* multiple character wildcard (matches any number of characters in a directory/file name)
... hierarchical wildcard (matches any number of hierarchical directories)
.. specifies the parent directory
. specifies the directory containing the lib.map
Paths that end in / shall include all files in the specified directory. Identical to /*.
Paths that do not begin with / are relative to the directory in which the current lib.map file
is located.
2) The paths ./*.v and *.v are identical, and both specify all files with a .v suffix in the current
directory

alaindargelas:
To be noted, Library/Config file parsing should not happen anywhere but in the main thread, when processing the project command line.

alaindargelas:
Have you noticed actual runtime issues?

hs-apotell:
> Please propose another implementation.

Will do.

> To be noted, Library/Config file parsing should not happen anywhere but in the main thread, when processing the project command line.

That might be true within the scope of the Surelog code but you can't draw that line for all the applications using Surelog. This function is affecting the state of entire application. I could envision cases where I invoke the Surelog parser on its own thread, especially, true for interactive UI applications.

> Have you noticed actual runtime issues?

No, not yet. I just happen to look thru' other function implementations when working on another task in that file. That said, I hope that is not a criteria to fix something that has been identified to be a potential issue. This is an implementation with side-effects which aren't explicit from its function name. Plus, this code is likely to have permission issues as well when attempting to change working directory to root and also stepping into other user's home directory.

hs-apotell:
@alaindargelas I have a working version of the collectFiles and I was hoping to compare the results of my implementation with yours. Ran into some problems along the way and following up to ensure my understanding is correct (and its a bug in current implementation).

[...]

PR #479, opened 2020-06-05
Issue #410: Implementation of FileUtils::collectFiles isn't thread safe

hs-apotell:
Issue #410
Implementation of FileUtils::collectFiles isn't thread safe.

Cause
The use of 'chdir' makes the implementation non-thread-safe.

Resolution
Reimplemented the function to work with filesystem and strings
only i.e. without any OS calls (except calls to scan for files)

### DIFF
--- a/src/Utils/FileUtils.cpp
+++ b/src/Utils/FileUtils.cpp
@@ -155,114 +155,76 @@ std::vector<SymbolId> FileUtils::collectFiles(const std::string dirPath,
   return result;
 }
 
-std::vector<SymbolId> FileUtils::collectFilesRegexp(const std::string dirPath,
-                                                    const std::string regexp,
-                                                    SymbolTable* symbols) {
+std::vector<SymbolId> FileUtils::collectFiles(const std::string& pathSpec,
+                                              SymbolTable* const symbols) {
+  // ?   single character wildcard (matches any single character)
+  // *   multiple character wildcard (matches any number of characters in a
+  // directory/file name)
+  // ... hierarchical wildcard (matches any number of hierarchical directories)
+  // ..  specifies the parent directory
+  // .   specifies the directory containing the lib.map
+  // Paths that end in / shall include all files in the specified directory.
+  // Identical to / * Paths that do not begin with / are relative to the directory
+  // in which the current lib.map file is located.
+
   std::vector<SymbolId> result;
-  if (!fileIsDirectory(dirPath)) {
-    return result;
-  }
 
-  std::string newregexp;
-  for (unsigned int i = 0; i < regexp.size(); i++) {
-    if (regexp[i] == '.') {
-      newregexp += "\\";
-    }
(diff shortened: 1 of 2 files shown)


hs-apotell:
@alaindargelas Two reasons for the Presubmit tests (specifically, UnitLibrary) failing -
1. The directory paths on the builder and the log files committed to the repo are different.
On builder - `/tmpfs/src/github/alainmarcel-surelog/tests/UnitLibrary/`
In Repo - `/home/alain/Surelog/tests/UnitLibrary/`
2. Two library files `UnitLibrary/libwconfig/libw1/wsub.v` and `UnitLibrary/libwconfig/libw2/wsub.v` are being found twice with the new implementation. This is the an outlier and the only one that's causing the problem. Could you please confirm that the inputs are correct and the problem isn't in the source itself.

Here's a list of my test cases that I have verified for the new algorithm -
For both unix and windows, relative paths -
* folder/
* folder/*.v
* folder/*.vhd
* folder/fsm?.v
* ./folder/
* ./folder/*.v
* ./folder/*.vhd
* ./folder/fsm?.v
* ./.../*
* ./.../*.v
* ./.../fsm?.v
* ./.../fsm?.vhd

Windows specific rooted tests -
* D:/Projects/Surelog/TestModules/
* D:/Projects/Surelog/TestModules/*.v
* D:/Projects/Surelog/TestModules/fsm?.v
* D:/Projects/Surelog/.../*
* D:/Projects/Surelog/.../*.v
* D:/Projects/Surelog/.../fsm?.v
* D:/Projects/Surelog/.../fsm?.vhd

Unix specific rooted tests -
* /Projects/Surelog/TestModules/
* /Projects/Surelog/TestModules/*.v
* /Projects/Surelog/TestModules/fsm?.v
* /Projects/Surelog/.../*
* /Projects/Surelog/.../*.v
* /Projects/Surelog/.../fsm?.v
* /Projects/Surelog/.../fsm?.vhd

Let me know if you want me to test the implementation with any other specific input.

alaindargelas:
The whole purpose of the Library concept is to support files with the same name and modules with the same name under different libraries (mapped to different directories). Disembuigation happens in the configuration stage.
The absolute path name is not the reason of the diff, absolute paths are ignored in the diffs.
The diff is due to an incorrect behavior of the new implementation or as you said the input might be misinterpreted. But this testcase was carefully crafted with the intention described above.
I'll check the new implementation before merging.

alaindargelas:
"Two library files UnitLibrary/libwconfig/libw1/wsub.v and UnitLibrary/libwconfig/libw2/wsub.v are being found twice with the new implementation."
That is the source f the problem, the input is correct, somehow the 

Somewhat the directives in UnitLibrary/libwconfig/libw.map 
library libw1 libw1/;
library libw2 libw2/;
are searching both directories as opposed to each one only searching their respective dirs.

PR #535, opened 2020-06-24
Change a few places that can use const references.

hzeller:
Using const references improves readability as it is easier for
a developer to see that nothing is modified.
(It also opens of course optimization opportunities, but well
 self-documenting code is the main motivation).

Also in this change a few places that
  - use std::string_view where safe to do
  - make some member variables 'const' where they are only
    set in the constructor and never modified.
  - make methods that don't modify internals 'const'.
-


hzeller:
Looks like ready to merge (only the broken Mac build is broken).

PR #538, opened 2020-06-24
SymbolTable: use const references and const methods where possible.

hzeller:
Also IWYU: remove some unnecessary includes in SymbolTable, which then
uncovered some other places which relied on that. Fixed them.
    
In the touched files: always include the header of the implementation
(so foo.h included in foo.cpp) at the very first: this should always
work cleanly and will find accidental dependencies on other headers.


hzeller:
Looks like ready to merge (only the broken Mac build is broken).

PR #540, opened 2020-06-24
Make dependencies work better in Makefile.

hzeller:
Some targets were only working if other targets had been
called before, e.g. regression required to have run the release
before. These dependencies are now better captured in Makefile
rules.

Also: make clean now just removes the build directory, instead
of first calling make clean and then removing the directory anyway.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,46 +9,42 @@ endif
 
 PREFIX ?= /usr/local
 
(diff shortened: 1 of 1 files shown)


hzeller:
Ready to merge. Only the failing Mac test is failing.

PR #548, opened 2020-06-25
Fix old TODO: don't use globbing to find files to compile.

hzeller:
This prevents cmake seing if there are changes etc.

This is also as recommended by CMake documentation:
  https://cmake.org/cmake/help/v3.15/command/file.html#filesystem


hzeller:
Also, sorting all the files so that they're easier to find.

hzeller:
Ready to merge. Only the good ol' Mac test that is failing anyway is failing.

PR #549, opened 2020-06-25
Remove logfile that looks like it was accidentally checked in.

hzeller:
It creates a lot of churn as it shows up as 'changed' whenever
a local test is run.


alaindargelas:
BTW, the Windows test does nothing

PR #550, opened 2020-06-25
Update UHDM submodule; disable submodule test in Surelog.

hzeller:


hzeller:
mmh, this looks funny: https://travis-ci.com/github/alainmarcel/Surelog/builds/173088752

`FAILDUMP` is that because it would try to run a dump binary ? The `dump` binary is part of the UHDM test, which are now disabled, so this would explain this. Ideally, the `dump` should actually be a binary distributed with UHDM as binary (because it is very useful in its own right), and then derive a test from that.

Let me enable the tests for now here.

hzeller:
Once https://github.com/alainmarcel/UHDM/pull/271 is in, I can make this one work.

PR #551, opened 2020-06-26
Avoid double operation in SymbolTable; simplify.

hzeller:
Replace insert-if-not-available with canonical C++ idiom.

When inserting into the SymbolTable, we first did a look-up, then
insert if we didn't find anything. Using the return value of
the container insert() method, we can do two things in one.

The UnitTests verified that things work the same as before.


hzeller:
Also, makes things more compact.
The `getId()` didn't change, just made more readable and compact by using the `?:` operator.

PR #553, opened 2020-06-26
Avoid virtual method table for classes that don't need it.

hzeller:
Virtual class instances require an additional pointer stored for
each instance, which uses unnecessary memory.
Found the classes in question by marking them final and see if there
are any complaints about derived classes not being able to
extend due to the finalness. For the ones without complaint, removed
the virtual destructor. Left them final, so if extensibility is needed
in the future, it will be intentional.

While at it, made methods, parameters and return values const where
possible. Fixed trailing whitespace.

In this pull request, mostly only addressed src/Design/*


hzeller:
Ready to merge.

PR #570, opened 2020-07-01
Cache: remove unnecessary virtual. Typedef a particular long type.

hzeller:
* SURELOG::Cache is extended, but does not have a virtual method.
   Change the virtual destructor to plain.
 * Make all methods in Cache protected, as they just serve as
   utilities in derived classes without being publically accessed.
 * use a using-clauses for a particular long flatbuffer vector of
   offsets of things.
 * Fix some indentation; remove unnecessary semicolon at end of
   namespace.


hzeller:
(Travis build was fine, but then timed out. Restarted it, but otherwise looks good)

PR #572, opened 2020-07-02
First step towards const correctness. Code improvement, no change in …

hzeller:
…functionality.

Mostly starting with FileContent, making a bunch of getters const,
add getters calld mutableFoo() if they return a modifyiable value.

Signed-off-by: Henner Zeller <h.zeller@acm.org>

PR #580, opened 2020-07-05
First step towards const correctness.

hzeller:
Code improvement, no change in functionality.

Mostly starting with FileContent, making a bunch of getters const,
add getters calld mutableFoo() if they return a modifyiable value.

This then triggered a cascade of things that then also could be
improved.


hzeller:
Re-do of #572 applying cleanly to current head.

PR #604, opened 2020-07-12
Antlr4 getEpsilonTarget() non shared_ptr allocation Experiment. Not for submit.

hzeller:
This is an experiment to see if performance in the central
loop thing gets better when we don't allocate a shared pointer
with every ATNConfig, but create and pass by value (with
some consideration returning a non-object using std::optional).

Only when we actually continue fromn there, the value is
allocated as shared ptr as this is how all the rest works...

There is one place where the pointer is inserted into a
set to limit recursion. This is done at a time when we have not
created the shared pointer yet, so just the hash of the value is
used.

Long story short: this does not change anything, it even gets
slightly slower.

But it might be a good as an failed example or starting point
for something more.

### DIFF
--- a/third_party/antlr4/runtime/Cpp/runtime/CMakeLists.txt
+++ b/third_party/antlr4/runtime/Cpp/runtime/CMakeLists.txt
@@ -1,3 +1,4 @@
+set(CMAKE_CXX_STANDARD 17)
 
 include_directories(
   ${PROJECT_SOURCE_DIR}/runtime/src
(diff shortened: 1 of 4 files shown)


hzeller:
I compared times running all the sv-tests, and it actually  looks like this version is not doing any better, probably even slightly worse. So this is mostly an FYI pull request, possibly useful for Octavian to also look at.

Baseline:

size-time-baseline

With this code:

size-time-with-atnconfig

.. overlay

size-time-compare

hzeller:
(graphs created using the improved csv output in https://github.com/SymbiFlow/sv-tests/pull/929 )

PR #613, opened 2020-07-13
TimeInfo: use enum class for enums and direct member initialization.

hzeller:
For type-safety and readability.
We now have to explicitly cast them to the flatbuffer uint16_t cache
values, but this make that transition explicit.


mithro:
Should CapnProto learn about C++ enums in the future?

hzeller:
(In this particular case, it is flatbuffers).

Low priority. I think it is ok to do casts on the serialization barriers.

PR #618, opened 2020-07-14
Remove unnecessary virtual destructor from ConfigSet.

hzeller:
Also change the accessor names to more clearly document that
we return a mutable config here. This will help later to find
all the places that can be re-considered to constify.

### DIFF
--- a/src/Config/ConfigSet.cpp
+++ b/src/Config/ConfigSet.cpp
@@ -20,16 +20,13 @@
  *
  * Created on February 10, 2018, 11:14 PM
  */
-#include "SourceCompile/SymbolTable.h"
-#include "Design/FileContent.h"
 #include "Config/ConfigSet.h"
+
 using namespace SURELOG;
 
-ConfigSet::~ConfigSet() {}
-
-Config* ConfigSet::getConfig(std::string configName) {
-  for (unsigned int i = 0; i < m_configs.size(); i++) {
-    if (m_configs[i].getName() == configName) return &m_configs[i];
+Config* ConfigSet::getMutableConfigByName(std::string_view configName) {
+  for (auto &config : m_configs) {
+    if (config.getName() == configName) return &config;
   }
-  return NULL;
+  return nullptr;
 }
--- a/src/Config/ConfigSet.h
+++ b/src/Config/ConfigSet.h
@@ -23,23 +23,26 @@
 
 #ifndef CONFIGSET_H
 #define CONFIGSET_H
+
 #include "Config/Config.h"
+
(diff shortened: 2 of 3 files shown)


alaindargelas:
Sorry, I don't like the Mutable adjectives polluting the assessor names.
I do lots of greps and autocomplete while developing, it sounds nice to add these adjectives when considering the code frozen, not so much while actively developing.
I typically also do code completion in the IDE having get... first letter autocomplete is much better than having to type getMutable...

alaindargelas:
I didn't pay attention to previous changelists adding this Mutable adjective, I'd rather undo them all than having more added.

hzeller:
Our goal is to ideally never hand out mutable pointers, because then it makes it really hard to make things const correct.
GetSomething() should always be const in the spirit, so returning something mutable, it should be marked as such, because it is completely unexpected if not. Right now we're at the beginning of seeing what can be made const before we then can go ahead and re-arranging things in a way that we can reduce the use of the Mutable-ness.

So this is why I suggest to have very obviously 'loud' annotation like Mutable.

In _particular_ for active development, these attributes are important, because then it helps making the right decisions. It doesn't help afterwards. Every single time one uses an object, the first instinct must be to use a const version of it unless really really needed. Because then the code flow for the developer and the people contributing to it is actually clear. It is very unclear right now.

alaindargelas:
UHDM is the API that should be const correct, and I think you made it so, that is the API external contributors use.

Surelog internal objects are modified in several passes, from different builders (Compilation, Elaboration, annotation for error messages...), adding more data in each pass.
Getting Incomplete objects (non const) and augmenting them with new data is happening all over the place.
There is no "fixed" data structure almost until we create the UHDM Datamodel at which point you can also consider the internal objects "read only", but until then most of them are getting mutated somewhere.

hzeller:
Yes, but this is why I am working on the improvement of Surelog that this is easier to follow and allows us to put together all the stateful things in a way that it works well. It is important to identify all the places where we modify a particular thing and where not. If it can't be done, then this is a maintenance problem. This is why I am working to get it into such a state where it is easy to reason which parts are modified where and not have a 'everything is modified everywhere' situation.

Getters almost never should hand out mutable state, because it means that they only act as holder objects and are not responsible for their internal state. It can be done, but we need to first simplify the baseline so that things can be refactored accordingly. This is the current stage I am working on - adding const where possible without changing the current structure, so that then it is even feasible to think how to restructure it.

alaindargelas:
Alright, I won't wiggle myself out of this one :-)

PR #620, opened 2020-07-14
Remove unnecessary virtual destructors in some classes never extended.

hzeller:
Also, remove copy constructors and make instance variables that are
only set in the constructor const.


hzeller:
Build failed, but this was just a Travis timeout. You can see that it ran most of the tests successfully before it timed out.

PR #636, opened 2020-07-15
Simplify Precompiled and remove unused virtual destructor.

hzeller:
Also move the constructors to private, as the only access is
through the singleton.


hzeller:
I don't know how the googletest thing ended up being in there. Let me do a fresh pull request.

PR #644, opened 2020-07-16
Remove global-namespace polluting using namespace in headers.

hzeller:
Limit Using declarations to compilation units.


hzeller:
The broken travis test is just a timeout. Don't know if/how I can restart it.

PR #645, opened 2020-07-16
Document and unit-test various ltrim()/rtrim() functions and leaf().

hzeller:
* Also found that the trim functions that take a character to trim
    don't exactly behave as the name implies. Not changed functionality
    here, just documenting for now.
  * Removed ltrimStat() which had a name that did not hint what
    it was supposed to be doing; it was also not used anywhere
    in the code base.

### DIFF
--- a/src/Utils/StringUtils.cpp
+++ b/src/Utils/StringUtils.cpp
@@ -254,16 +254,6 @@ std::string& StringUtils::ltrim(std::string& str, char c) {
   return str;
 }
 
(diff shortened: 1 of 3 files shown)


hzeller:
Looks like same problem again with travis timing out, but the other tests went through fine.

I wonder if in travis, we should try a moderate parallel two-core run of the regression-tests ? Or do we run out of memory ?

alaindargelas:
The Unisim test took us over the 50 mins threashold that Travis seems to have.

PR #649, opened 2020-07-17
Canonicalize filenames in regression logs (and fix whitespace in regression.tcl)

hzeller:
1) [First change](https://github.com/alainmarcel/Surelog/pull/649/commits/13711e4ab32f29fda64befff24427ada2772318c): No functionality change, just whitespace fixes in tcl (using 4 space for indentation, no tabs)
  2) [Second change](https://github.com/alainmarcel/Surelog/pull/649/commits/a8ea5700b4a1eee3358c587663023ba78ef8f79f): Canonicalize filenames mentioned in logfiles `/home/alain/Surelog/foo/bar` or `/home/hzeller/src/Surelog/foo/bar` -> `$(SURELOG_DIR)/foo/bar` so that we can create a diff that works on all developers' machines and also fixes #39 
 3) [Third change](https://github.com/alainmarcel/Surelog/pull/649/commits/a15f31e454c4cfdec9deb0302a99617013d663ad) canonicalize diagnostic time output, as this will also vary slightly from regression to regression and in particular on different machines.

PR #647, opened 2020-07-16
Run tests on travis with 2 cores, cutting CI times and avoiding timeouts.

hzeller:
Signed-off-by: Henner Zeller <h.zeller@acm.org>

PR #650, opened 2020-07-17
Append 's' to time outputs.

hzeller:
This makes it simpler to add time canonicalization similar to
filename canonicalization added in #649: we can confidently
latch on a regular expression to something [0-9]+.[0-9]{3}s or
[0-9]+.[0-9]{6}s.


hzeller:
Travis shows the usual build timeout, but that is fixed once #647 is merged.

PR #717, opened 2020-08-07
Emit empty lines for comments in inactive blocks.

### LINKED ISSUE #715: Preprocessing: comment lines in removed ifdef-region not counted
```
build/bin/surelog -parse third_party/tests/Earlgrey_0_1/src/lowrisc_prim_util_memload_0/rtl/prim_util_memload.sv
[INF:CM0023] Creating log file ./slpp_all/surelog.log.

[SNT:PA0207] third_party/tests/Earlgrey_0_1/src/lowrisc_prim_util_memload_0/rtl/prim_util_memload.sv:45: Syntax error: mismatched input 'initial' expecting <EOF>,
initial begin
^-- ./slpp_all/work/./third_party/tests/Earlgrey_0_1/src/lowrisc_prim_util_memload_0/rtl/prim_util_memload.sv:45:0:.
```

The error is reported in line 45, but it actually in [line 49 of that file](https://github.com/alainmarcel/Surelog/blob/master/third_party/tests/Earlgrey_0_1/src/lowrisc_prim_util_memload_0/rtl/prim_util_memload.sv#L49)

Comparing the original with the preprocessed file, where the ifdef-region is removed, we can see that not all lines from the input are actually emitted as empty lines in the output:
```
diff -u third_party/tests/Earlgrey_0_1/src/lowrisc_prim_util_memload_0/rtl/prim_util_memload.sv ./slpp_all/work/./third_party/tests/Earlgrey_0_1/src/lowrisc_prim_util_memload_0/rtl/prim_util_memload.sv | awk '/^-/ { minus++; } /^+/ { plus++; } END {printf("removed %d added %d\n", minus, plus); }'
removed 30 added 26
```

Looking at the original code, it contains exactly **4** lines of comments between the ifdef/endif....

Hyopthesis: these don't make it as empty output and are thus not counted.

Test: If we remove these comments, the error reporting is on the correct line:
```
sed 's|//.*||g' < third_party/tests/Earlgrey_0_1/src/lowrisc_prim_util_memload_0/rtl/prim_util_memload.sv > no-comments.sv

# Now let's run surelog on the same file with these comments removed
build/bin/surelog -parse no-comments.sv
[INF:CM0023] Creating log file ./slpp_all/surelog.log.

[SNT:PA0207] no-comments.sv:49: Syntax error: mismatched input 'initial' expecting <EOF>,
initial begin
^-- ./slpp_all/work/./no-comments.sv:49:0:.
```

Now, the error is reported in the correct line.
Conclusion: comment lines inside removed ifdef blocks are not emitted as empty lines so the line number reporting is off from then on.

(This off line-number reporting was observed here https://symbiflow.github.io/sv-tests/#Surelog|ibex|ibex which became the base for the easy-to-reproduce case above)

hzeller:
Fixes #715

### DIFF
--- a/src/SourceCompile/SV3_1aPpTreeShapeListener.cpp
+++ b/src/SourceCompile/SV3_1aPpTreeShapeListener.cpp
@@ -131,6 +131,8 @@ void SV3_1aPpTreeShapeListener::exitComments(SV3_1aPpParser::CommentsContext * c
       } else if (ctx->One_line_comment()) {
         addVObject (ctx, ctx->One_line_comment()->getText(), VObjectType::slComments);
       }
+    } else {
+      addLineFiller(ctx);
     }
   }
 }


hzeller:
mmh, the whole logfile diffing is still not optimal, this is why the tests above fail.
I'll have to look at that.

alaindargelas:
This is actually correct, you should just have updated the logs:
cd build

You eliminated a bunch of messages that complain about pointing to an empty line.
I'll merge and update the logs.

PR #761, opened 2020-09-01
Modifying when the Cache is set up so -nocache flag is checked first

LoJeff:


alaindargelas:
@LoJeff to create a test:
cd Surelog/tests
mkdir <testName>
- Add a dut.sv, testName.sl file (See other tests as example)
 cd Surelog/build
../tests/regression.tcl test=<testName>
../tests/regression.tcl test=<testName> update

- git add tests/testName
- create a pull request

LoJeff:
@alaindargelas 
Is there an infrastructure currently for the tests to check if a file is generated or not?

If not I see two ways to approach this:
1) Print an info message when cache is not generated
2) Add a file checking functionality to the test infrastructure

alaindargelas:
Please do print an info message and update the log file.That should be enough.
    On Thursday, October 1, 2020, 11:09:36 AM PDT, Jeff Lo <notifications@github.com> wrote:  
 
 


@alaindargelas
Is there an infrastructure currently for the tests to check if a file is generated or not?

If not I see two ways to approach this:
   
   - Print an info message when cache is not generated
   - Add a file checking functionality to the test infrastructure

PR #787, opened 2020-09-17
Delegate method handling for this, super etc.

rkapuscik:
Compile method calls for these statements in `compileComplexFuncCall`,
which already accounts for the keywords. This prevents leaving them as
plain `vpiSysFuncCall` in UHDM.

Many logs for test cases still need to be updated, but I'd like to have your opinion on this approach. The `compileComplexFuncCall` needs to be expanded as well before merging this, but at least the code responsible for the calls would be kept in one place.

### DIFF
--- a/src/DesignCompile/CompileHelper.cpp
+++ b/src/DesignCompile/CompileHelper.cpp
@@ -1815,18 +1815,22 @@ UHDM::tf_call* CompileHelper::compileTfCall(DesignComponent* component, const Fi
     name = fC->SymName(fC->Child(dollar_or_string));
   } else if (leaf_type == slImplicit_class_handle) {
     NodeId handle = fC->Child(dollar_or_string);
-    if (fC->Type(handle) == slSuper_keyword) {
-      name = "super.";
-    } else if (fC->Type(handle) == slThis_keyword) {
-      name = "this.";
+    if (fC->Type(handle) == slSuper_keyword ||
+        fC->Type(handle) == slThis_keyword ||
+        fC->Type(handle) == slThis_dot_super) {
+      return (tf_call*) compileComplexFuncCall(component,
+                                               fC,
+                                               Tf_call_stmt,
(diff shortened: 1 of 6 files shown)


alaindargelas:
Got it, please continue on this path, looks good!

alaindargelas:
Please update the logs.

alaindargelas:
What happened to the handling of:

  name = "super.";

and the likes,

From the changes, I see that you removed it in one place, but I don't see where is this handled anymore.

src/DesignCompile/CompileHelper.cpp
```diff
@@ -1815,18 +1823,22 @@ UHDM::tf_call* CompileHelper::compileTfCall(DesignComponent* component, const Fi
     name = fC->SymName(fC->Child(dollar_or_string));
   } else if (leaf_type == slImplicit_class_handle) {
     NodeId handle = fC->Child(dollar_or_string);
-    if (fC->Type(handle) == slSuper_keyword) {
-      name = "super.";
```
alaindargelas, line 1819: Where is this happening now?

alaindargelas:
You have to merge #797 to compute the correct log files

src/DesignCompile/CompileHelper.cpp
rkapuscik: It's handled here: https://github.com/alainmarcel/Surelog/blob/master/src/DesignCompile/CompileExpression.cpp#L2500, then is actually compiled in these cases: https://github.com/alainmarcel/Surelog/blob/master/src/DesignCompile/CompileExpression.cpp#L1246.

rkapuscik:
> You have to merge #797 to compute the correct log files

Thanks for the heads up, the branch is now rebased and logs are updated.
I think this is ready to be merged now, though there is still work to be done in `compileExpression` to improve prefix compilation for methods.

PR #839, opened 2020-10-14
Add attribute record on source_text top element

Thomasb81:
Add attributes on module, class, package, interface, primitive
Update Attributes test to reflect the enhancement


Thomasb81:
So much diff on src/DesignCompile/UhdmWriter.cpp were not expected ...

alaindargelas:
@Thomasb81, looks good!
FYI, the last thing you were missing to get a clean regression was to update the log.
A quick way to update all logs:
cd build

Of course check the diffs prior to update, but they also can be checked here in the PR along with the source change.

Issue #631, opened 2020-07-15
Add Travis (and Kokoro?) build for Windows


mithro:
_Originally posted by @alaindargelas in https://github.com/alainmarcel/Surelog/issues/239#issuecomment-658534307_

> Windows support achieved, Thanks @hs-apotell!
> Do we need a travis Windows build?

We should have CI running on Windows so that we don't break the Windows build accidentally.

hs-apotell:
Absolutely, Yes. I started working on it and then got stuck with the requirement for tcl for windows. It was a while back and might have gotten stale.

Related issue - https://github.com/alainmarcel/UHDM/issues/229

hs-apotell:
We have a basic windows CI support working. However, we are running into some issues.

1. We have about 16 test failures. Attached are the logs for failing tests and an excel document summarizing them.
2. Our builds keep timing out on travis. This is happening for Unix builds also - which is more troubling because the same builds under @alaindargelas account seem to be finishing much faster.

Any assistance in identifying the root causes and/or fixing these issues or sharing any pointers helping us do it is highly appreciated.
[logdir.tar.gz](https://github.com/alainmarcel/Surelog/files/5383765/logdir.tar.gz)
[surelog_test_diff.xlsx](https://github.com/alainmarcel/Surelog/files/5383766/surelog_test_diff.xlsx)

umarcor:
I tried creating a GitHub Actions workflow: https://github.com/umarcor/Surelog/blob/ci/add-gha-workflow/.github/workflows/test.yml

However, I'm working on Windows, and it seems that this repository contains several files which have the same name but with different capitalisation: https://github.com/umarcor/Surelog/commit/752b17dfd2d9c0a014f50408036375ffdc49e54f. That's unfortunate, because it produces many conflicts when trying to rebase commits.

On the other hand, when cmake is executed, the following error is produced:

```
 CMake Error at third_party/UHDM/CMakeLists.txt:29 (add_subdirectory):
  The source directory

    /home/runner/work/Surelog/Surelog/third_party/UHDM/third_party/capnproto

  does not contain a CMakeLists.txt file.


CMake Error at third_party/UHDM/CMakeLists.txt:249 (install):
  install TARGETS given target "capnp" which does not exist.
```

Hence, it seems that an issue is coming from a submodule.

umarcor:
[...]

PR #853, opened 2020-10-18
Issue #631: Regression test comparison failing

hs-apotell:
Issue #631

Issue:
Regression tests comparison where failing (false negatives) on Windows

Root Cause:
UHDM coverage logic wasn't compensaring for LF vs. CRLF

Solution:
Reimeplemented the logic to split a buffer to its individual line.
Also, before comparing replaced all LF and CRLF with empty string.

P.S. This change requires updating golden images for a few tests.


hs-apotell:
@alaindargelas This change requires updating the golden logs for a few regression tests. The regression tests are expected to fail for this change. I just wasn't sure what the process was to update the golden logs.

alaindargelas:
@hs-apotell I'll do the update.
Net time, simply do after running the regression:
cd build

PR #847, opened 2020-10-17
ci/test: add windows job

umarcor:
As discussed in #631, this is a placeholder for adding a Windows job to the GitHub Actions workflow contributed in #845. Currently the build sources in this repo don't support building on Windows out of the box (#844). Hence, this is to be kept as a draft meanwhile. Contributors willing to test Windows support in CI can checkout this branch, apply changes and push to their fork.

PR #863, opened 2020-10-19
Restore behavior for builds and regression to run in parallel

hs-apotell:
Changed all build commands to run in parallel. Gives a few minutes improvements over past results. Here's a run on GHA -
https://github.com/anadishukla/Surelog/actions/runs/314676434

@umarcor We should take out the -j option in GHA file. The Makefile takes care of making it parallel.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,10 +13,10 @@ endif
 PREFIX ?= /usr/local
 
 release: run-cmake-release
-	cmake --build build
+	cmake --build build -j $(CPU_CORES)
 
 debug: run-cmake-debug
-	cmake --build dbuild
+	cmake --build dbuild -j $(CPU_CORES)
 
 run-cmake-release:
 	mkdir -p build/tests dist
@@ -27,7 +27,7 @@ run-cmake-debug:
 	cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_INSTALL_PREFIX=$(PREFIX) -S . -B dbuild
 
 test/unittest: run-cmake-release
-	cmake --build build --target UnitTests
+	cmake --build build --target UnitTests -j $(CPU_CORES)
 	pushd build; ctest --output-on-failure; popd
 
 test/regression: run-cmake-release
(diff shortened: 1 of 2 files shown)


umarcor:
Done #847.

hs-apotell:
> 
> 
> Done #847.

It applies to Linux builds as well. The change in this PR finishes faster than most other builds so far. There shouldn't be any difference between Linux and Windows Build and Test steps (except for the use of sudo on Linux for install command)

https://github.com/alainmarcel/Surelog/actions/runs/314724614 - finished in 40 mins.
The fastest in previous 10 successful builds has been 49 mins.

alaindargelas:
@hs-apotell Can you double check that 
"make regression" does indeed run with as many CPUs are there are on the machine?
Last I tried your change, it was limited to 2 CPUs, that's why I undid you change.
I have a 24 CPU server and the regression runs in 4 minutes. With you change it ran in 20 minutes last I tried.
That's why I'm asking again, have you validated it with say 4 or 8 cores or more?

alaindargelas:
Note that the builds and regression where running in parallel. The feature was lost with some of your previous checkins.

hs-apotell:
I have a 8 core machine and I see 8 parallel surelog processes running during regression (and 8 parallel cl processes during build) with this change. In fact, the build on GHA also was faster by around 10 mins.

P.S. I did edit the Makefile to add `-j <n>` option to all build commands. This is documented in [here](https://cmake.org/cmake/help/v3.12/manual/cmake.1.html#build-tool-mode)

alaindargelas:
What I meant is that before the recent switch to cmake for the sub steps of the Makefile, using make, all was parallel.I'll merge tomorrow. 

    On Monday, October 19, 2020, 12:49:29 AM PDT, HS <notifications@github.com> wrote:  
 
 


I have a 8 core machine and I see 8 parallel surelog processes running during regression (and 8 parallel cl processes during build) with this change. In fact, the build on GHA also was faster by around 10 mins.

P.S. I did edit the Makefile to add -j <n> option to all build commands. This is documented in here

Issue #910, opened 2020-11-17
Windows build fails (java not found)


tgorochowik:
Example:
https://github.com/alainmarcel/Surelog/runs/1412535103?check_suite_focus=true#step:9:304
```
...
---------- Parser generation...
/bin/sh: java: command not found
[187/479] Building CXX object third_party/UHDM/third_party/capnproto/c++/src/kj/CMakeFiles/kj-async.dir/async.c++.o
ninja: build stopped: subcommand failed.
make: *** [Makefile:16: release] Error 1
Error: Process completed with exit code 2.
```

The issue seems to be a hard-coded java path here:
https://github.com/alainmarcel/Surelog/actions/runs/368229903/workflow#L108
```
export JAVA_HOME=/c/hostedtoolcache/windows/jre/8.0.272/x64
```

We need to figure out the actual path dynamically as the environment is regularly updated and this path changes.

Additionally, the checks for compiler versions do not seem to work:
https://github.com/alainmarcel/Surelog/actions/runs/368229903/workflow#L122-L134
```
    - name: Show shell configuration
      run: |
        export PATH=$SWIG_DIR:$JAVA_HOME/bin:$PATH
        env
        where cmake && cmake --version
        where make && make --version
        where swig && swig -version
        where java && java -version
        where python && python --version
        where ninja && ninja --version
        where tclsh && echo 'puts [info patchlevel];exit 0' | tclsh
        where $CC && $CC --version
        where $CXX && $CXX --version
```

I am not sure what the reason is, but even if there is no java, the pipeline just continues, as can be seen here:
https://github.com/alainmarcel/Surelog/runs/1412535103?check_suite_focus=true#step:8:193
(you can see the version printed for `swig`, and then immediately for `python`).

tgorochowik:
@hs-apotell would you mind taking a look at this?

PR #911, opened 2020-11-18
Issue #910: Fix java path and make Java mandatory

hs-apotell:
Issue #910 : Java not found

Resolution:
* Make JAVA_HOME environment variable be dependent on variables setup by configuration action
* Make Java mandatory so build fails when java isn't found

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -105,7 +105,7 @@ jobs:
     - name: Configure shell environment variables
       run: |
         export SWIG_DIR=/c/ProgramData/chocolatey/lib/swig/tools/install/swigwin-3.0.12
-        export JAVA_HOME=/c/hostedtoolcache/windows/jre/8.0.272/x64
+        export JAVA_HOME=`cygpath -u $JAVA_HOME_8_0_275_X64`
         export CWD=`pwd`
 
         echo "SWIG_DIR=$SWIG_DIR" >> $GITHUB_ENV
@@ -122,6 +122,7 @@ jobs:
     - name: Show shell configuration
       run: |
         export PATH=$SWIG_DIR:$JAVA_HOME/bin:$PATH
+        export PATH=$SWIG_DIR:$PATH
         env
         where cmake && cmake --version
         where make && make --version
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -79,7 +79,7 @@ jobs:
     - name: Setup Java
       uses: actions/setup-java@v1
       with:
-        java-version: 1.8.275
+        java-version: 1.8.0.275
         java-package: jre
         architecture: x64
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -79,7 +79,7 @@ jobs:
     - name: Setup Java
       uses: actions/setup-java@v1
       with:
-        java-version: 1.8
+        java-version: 1.8.275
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -105,7 +105,7 @@ jobs:
     - name: Configure shell environment variables
       run: |
         export SWIG_DIR=/c/ProgramData/chocolatey/lib/swig/tools/install/swigwin-3.0.12
-        export JAVA_HOME=/c/hostedtoolcache/windows/jre/8.0.272/x64
+        export JAVA_HOME=`cygpath -u $JAVA_HOME_8_0_275_X64`
```
tgorochowik, line 108: I don't think it will resolve the issue, the version is still hardcoded here (so it will break again when the env is updated)

tgorochowik (commented):
Thank you for a very quick response!

tgorochowik:
Now that I think about it I see that java is not the only hardcoded PATH, the same is done for swig for example.

Maybe instead of finding the PATHs dynamically we could freeze the version of the environment so it doesn't get updated without our knowledge?

See here:
https://github.com/alainmarcel/Surelog/blob/ef488c691c7ff48398b9a4e7eb58a8fa5ea4543a/.github/workflows/main.yml#L151

Is there a way to freeze the env version instead of using `-latest`?

tgorochowik:
This might be helpful: https://github.com/actions/virtual-environments

.github/workflows/main.yml
hs-apotell: No it won't because we are explicitly asking for Java 1.8 version above. The minor version might change which is what this change is accounting for.

hs-apotell:
> 
> 
> Now that I think about it I see that java is not the only hardcoded PATH, the same is done for swig for example.
> 
> Maybe instead of finding the PATHs dynamically we could freeze the version of the environment so it doesn't get updated without our knowledge?
> 
> See here:
> 
> https://github.com/alainmarcel/Surelog/blob/ef488c691c7ff48398b9a4e7eb58a8fa5ea4543a/.github/workflows/main.yml#L151
> 
> Is there a way to freeze the env version instead of using `-latest`?

Swig and Java are the only hard-coded paths, both of which we are installing it explicitly as part of the workflow. Swig is installed using chocolaty and those packages are locked. They will always install at the same location. Java is installed/configured using the action, which can technically change the path because of minor version change and that is what I have accounted for with this change.

.github/workflows/main.yml
tgorochowik: I am not sure I understand. What happens when the java package is updated in the runtime environment again? It looks as though you are still asking for a specific java version (which is likely to be updated again in the future)

tgorochowik: According to https://github.com/actions/setup-java/blob/d202f5dbf7256730fb690ec59f6381650114feb2/README.md
we should also be able to lock java to a specific semver here:
https://github.com/alainmarcel/Surelog/blob/ef488c691c7ff48398b9a4e7eb58a8fa5ea4543a/.github/workflows/main.yml#L82

Maybe we should also change it here to `8.0.275` to make the package doesn't get updated without our knowledge.

hs-apotell: You are right. Updated the requested version to include the minor as well.

tgorochowik (approved):
The CI passes, thanks for you work!

PR #912, opened 2020-11-20
Remove extra ' " ' from STRING

kamilrakoczy:
This PR removes ' " ' from STRING field.


kamilrakoczy:
I see ``windows-msvc`` tests are failing, but it is not related with this PR. Only test that is failing (on run in this PR) is ``YosysSmallBoom``, but looking at antmicro branch (https://github.com/antmicro/Surelog/runs/1430434302) ``YosysSmallBoom`` passes, but ``Compl1001`` and ``BuildUVMPkg`` is failing.

I also did extra run for current master (https://github.com/antmicro/Surelog/runs/1431249448) and there ``YosysOpenSparc`` test failed, so I suspect, that there is some problems with ``windows-msvc`` and random tests are sometimes failing.

Issue #972, opened 2020-12-17
Wrong UHDM for nested struct/enum access


kamilrakoczy:
Test:
```
module top(input a, output b);
typedef enum logic [1:0] {
  PMP_MODE_OFF   = 2'b00,
  PMP_MODE_TOR   = 2'b01,
  PMP_MODE_NA4   = 2'b10,
  PMP_MODE_NAPOT = 2'b11
} pmp_cfg_enum;

typedef struct packed {
        logic a;
        logic b;
} pmp_cfg_struct;

typedef struct packed {
  pmp_cfg_enum mode;
  pmp_cfg_struct struct_var;
} pmp_cfg_t;

pmp_cfg_t      csr_pmp_cfg_i  [4];
pmp_cfg_t      csr_pmp_cfg_ie;
pmp_cfg_enum enum_test;

assign enum_test = PMP_MODE_OFF;

logic t = enum_test[0];

pmp_cfg_enum first = csr_pmp_cfg_i[0].mode;
logic second = csr_pmp_cfg_ie.mode[0];
logic third = csr_pmp_cfg_ie.struct_var.a;

assign csr_pmp_cfg_ie.mode = PMP_MODE_OFF;
endmodule
```
There is missing ``.a`` in nested struct access:
```
   |vpiContAssign:
   \_cont_assign: , line:24, parent:work@top
     |vpiRhs:
     \_ref_obj: (csr_pmp_cfg_ie.struct_var), line:24
       |vpiName:csr_pmp_cfg_ie.struct_var
     |vpiLhs:
     \_logic_net: (work@top.third), line:24, parent:work@top
       |vpiName:third
       |vpiFullName:work@top.third
       |vpiNetType:36
```
There is missing ``[0]`` when selecting part of enum:
```
   |vpiContAssign:          
   \_cont_assign: , line:23, parent:work@top
     |vpiRhs:    
     \_ref_obj: (csr_pmp_cfg_ie.mode), line:23                      
       |vpiName:csr_pmp_cfg_ie.mode
     |vpiLhs:                                   
     \_logic_net: (work@top.second), line:23, parent:work@top
       |vpiName:second                               
       |vpiFullName:work@top.second
       |vpiNetType:36
```
and after fixing problem with missing ``[0]``, accessing member of array selects bit of that member (instead of ``csr_pmp_cfg_i[0].mode`` it gives ``csr_pmp_cfg_i.mode[0]``):
```
[...]

PR #981, opened 2020-12-22
wip:hierarchical path

alaindargelas:


alaindargelas:
This changelist introduces a new UHDM object which is not present in the Standard Model: hier_path.
hier_path models hierarchical paths as a list of items, each item models using the Standard objects (obj_ref, bit_select...) the corresponding member of the path.
Example
assign b = c[0].d;
c[0].d will be modeled as a hier_path with 2 items, a bit select for "c[0]" and a ref_obj for "d".

The Standard Object model does not allow to model complex paths, instead VPI presents a bit blasted (bit select...) of the object pointed to by the path.
Since we want to present a folded model whenever possible, we introduced the hier_path object.

Changes to the model: 
- there is no binding performed of the different objects in the path (previously, in some simple cases bit_select or ref_obj where created to model the path and where bound, when possible to objects).
- Interface path, modport path are no longer bound for the same reason above.

We will work on a proper binding in Surelog in the near future, but the client applications can also perform that task as all the symbols are exported in UHDM.

In many cases, before this change the previous modeling of these hierarchical paths was actually incorrect in Surelog.
 
 Special notification for: @hs-apotell, @DamienLeBars, @bsp13

alaindargelas:
Fixes #972

alaindargelas:
Notify @Thomasb81

PR #1033, opened 2021-01-20
Add -noparametersubstitution parameter and remove -yosys/-verilator

kamilrakoczy:
This PR removes ``-yosys`` and ``-verilator`` parameters and instead introduces ``-noparametersubstitution``

CC: @tgorochowik

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,8 +87,6 @@ For more build/test options and system requirements for building see
    -nocomp               Turns off Compilation & Elaboration
    -noelab               Turns off Elaboration
    -elabuhdm             Forces UHDM/VPI Full Elaboration, default is the Folded Model
-   -verilator            Creates Verilator-friendly UHDM db (Fixes for Verilator limitations)
-   -yosys                Creates Yosys-friendly UHDM db (Fixes for Yosys limitations)
    -batch <batch.txt>    Runs all the tests specified in the file in batch mode. Tests are expressed as one full command line per line.
    -pythonlistener       Enables the Parser Python Listener
    -pythonlistenerfile <script.py> Specifies the AST python listener file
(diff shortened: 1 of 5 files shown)


alaindargelas:
Please implement Henner's suggestion.

tgorochowik:
Pasting the suggestion from @hzeller for reference:

> Can we maybe just name the specific features to be enabled and
> disabled and make them configurable with flags and then have something
> like --enable-feature=foo,bar --disable-feature=baz ?
> Then we can have sets of yosys or verilator specific configurations.
> And the code is much cleaner and we have a very good documentation of
> what is different for the various tools depending on the flag-set.

An additional suggestion was to see if we could wrap the output-changing parameters into a single struct so it can be easily filled externally in the future (for example for reading the config from a file)

alaindargelas:
Please make sure that struct is a data member of the command line class and accessible from it. Do not pass that struct around in all the function calls. From the code it is needed to be queried there is a always a back pointer to a series of objects that carry different contexts and the top most is the command line.

  On Tue, Jan 26, 2021 at 7:29, Tomasz Gorochowik<notifications@github.com> wrote:   


Pasting the suggestion from @hzeller for reference:


Can we maybe just name the specific features to be enabled and
disabled and make them configurable with flags and then have something
like --enable-feature=foo,bar --disable-feature=baz ?
Then we can have sets of yosys or verilator specific configurations.
And the code is much cleaner and we have a very good documentation of
what is different for the various tools depending on the flag-set.


An additional suggestion was to see if we could wrap the output-changing parameters into a single struct so it can be easily filled externally in the future (for example for reading the config from a file)

wtatarski:
@alaindargelas  
I added --enable-feature/--disable-feature option. Also I replaced `-yosys` and `-verilator` parameters with `parametersubstitution`, but actually after your last changes I wonder if this feature is still required, because I see that `-verilator` is not handled on master.

PR #1088, opened 2021-03-01
Make assignment operators blocking

rkapuscik:
According to IEEE 11.4.1
```
An assignment  operator  is  semantically  equivalent  to  a  blocking  assignment, 
with  the  exception  that  any left-hand index expression is only evaluated once. 
```
This affects assignment operators in for loop step statements, which were reported as nonblocking in UHDM.

### DIFF
--- a/src/DesignCompile/CompileHelper.cpp
+++ b/src/DesignCompile/CompileHelper.cpp
@@ -2328,13 +2328,16 @@ UHDM::assignment* CompileHelper::compileBlockingAssignment(DesignComponent* comp
   assignment* assign = s.MakeAssignment();
   UHDM::delay_control* delay_control = nullptr;
   if (Delay_or_event_control) {
-    delay_control = s.MakeDelay_control();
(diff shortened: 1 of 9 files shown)


alaindargelas:
@rkapuscik, please first create an issue for the problem you are seeing and include a testcase that shows the problem, 
I don't see a corresponding issue logged.

You should never do any type of string comparison on values, especially not with BAD_SYMBOL. After I see a case, I will propose a better solution.

rkapuscik:
I created an issue with a test case: #1092.

The comparison was supposed to create `vpiDelay` only if there is a valid value for the delay. Is there another way to check if the result of `fC->SymName()` is valid?

alaindargelas:
#1096 fixes this problem more correctly. The test on value was not required in the first place, the test on the Variable_lvalue early on in the function allowed the function code to follow the proper path in accordance with the Antlr grammar possibilities.

PR #1155, opened 2021-03-29
Fix setting parameter to 0 from command-line

kamilrakoczy:
Currently Surelog returns value ``48`` (ASCII ``0``) for parameters set using ``-Pparam=0``. This PR fixes this.

``strtoull`` is returning ``0`` if no valid conversion can be performed, but it is also returning ``0`` if it converts string to ``0``. Proper way of detecting if conversion was possible is comparison between pointer where comparison started and ended.

Signed-off-by: Kamil Rakoczy <krakoczy@antmicro.com>

### DIFF
--- a/src/Expression/ExprBuilder.cpp
+++ b/src/Expression/ExprBuilder.cpp
@@ -802,7 +802,9 @@ Value* ExprBuilder::fromVpiValue(const std::string& s) {
 Value* ExprBuilder::fromString(const std::string& value) {
   Value* val = nullptr;
   std::string sval;
-  if (strstr(value.c_str(), "'")) {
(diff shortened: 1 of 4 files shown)


kamilrakoczy:
@alaindargelas 
I'm believe it is not related with this PR, but it looks like there is some other problem with evaluation/overriding string parameters.
In test, where parameter is override with string, both ``GOOD_BLAH`` and  ``GOOD_0`` modules are created, but only ``GOOD_BLAH`` should be created: https://github.com/alainmarcel/Surelog/pull/1155/files#diff-2ae75e020b74d08ce5da83902fd4975b9c7a3d187c4c0e78ab59fecf620b0581R20

PR #1165, opened 2021-04-01
column info in UHDM

alaindargelas:


hs-apotell:
If I am reading this correctly, this doesn't provide the end line/column. It only provides the start line/column. Did you have something else in mind to get the missing information??

alaindargelas:
Correct, it does not provide end column, only start column.Getting the actual raw source content of the line (there is a function in Surelog that does that for error reporting) and knowing the start column, you should be able to do a simple token analysis to determine when the token (Keyword, name) ends for now.I might get another look at how to get the token end from antlr, but not soon. Other tasks are piling up.

  On Thu, Apr 1, 2021 at 1:01, ***@***.***> wrote:   


If I am reading this correctly, this doesn't provide the end line/column. It only provides the start line/column. Did you have something else in mind to get the missing information??

hs-apotell:
Our use case doesn't restrict with token identification/selection only. We want to be able to identify blocks also (think module definition, mathematical expressions, function bodies and such). We will need the end line/column information to make this useful. Just start line/column isn't enough.

alaindargelas:
Alright, still working on it.

  On Thu, Apr 1, 2021 at 21:24, ***@***.***> wrote:   


Our use case doesn't restrict with token identification/selection only. We want to be able to identify blocks also (think module definition, mathematical expressions, function bodies and such). We will need the end line/column information to make this useful. Just start line/column isn't enough.

alaindargelas:
I checked in the initial support for endline and endcolumn.
Right now the infrastructure is all there, but only the information for
module definition is filled in UHDM.
The Surelog internal AST (Cache) contains the information for all nodes.
Please help out by adding in UHDM the information for the objects you are
interested in.

hs-apotell:
Sure thing. Appreciate the prompt response in implementing this feature. It was long due.

PR #1257, opened 2021-05-05
Use external vpi_user.h from the single source in UHDM.

hzeller:
vpi_user.h was in both, UHDM and Surelog, and
they started to diverge (manual changes in UHDM).

In whole of Surelog, it was only used in one place;
use the version in UHDM.


hzeller:
msys2 CI error looks like it does not find the `diff` binary, unrelated to this change.
Linux CI diff looks unrelated.

hzeller:
CI issues look unrelated.

PR #1258, opened 2021-05-05
Move inclusion of process.h to the front.

hzeller:
According to the comment, this has to be always
before thread.h on Windows. Make sure it always is
once we sort the headers according to the style-guide.


hzeller:
I am not sure what the CI error is, but it does not look related.

hzeller:
CI issues look unrelated.

PR #1264, opened 2021-05-06
Add script to run clang format to be used in CI later.

hzeller:
This script runs clang-format on all non-generated
code, that will automatically format the code according
to the best practices and conventions in the style guide
https://google.github.io/styleguide/cppguide.html

Adding this script does not do anything yet. Steps to
follow:
  1) run it once to format existing code as a baseline.
  2) Add this as a script that runs in the CI.

### DIFF
--- a/.github/bin/run-clang-format.sh
+++ b/.github/bin/run-clang-format.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+# Copyright 2021 Alain Dargelas
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


hzeller:
It is probably best if you run this script when you don't have any changes ongoing, and then submit.
Since there is a lot that will slightly move around it would otherwise be likely to create conflicts.

Or I can also send a pull request if you like.

alaindargelas:
Let me try

alaindargelas:
Please submit a put request to enable the script in the ci.

PR #1272, opened 2021-05-07
Add code formatting check to CI

hzeller:


hzeller:
Tested that it worked.

PR #1280, opened 2021-05-10
Fix path handling on windows and include LogListener for installation

hs-apotell:
* Fix handling of absolute paths with forward slashes on windows.
* Include LogListener.h for install


alaindargelas:
@hs-apotell, please fix this warning. Thanks

PR #1290, opened 2021-05-13
Fix chandle type handling

RRozak:
`chandle` type was parsed as `logic_net`.  This commit fixes it.


alaindargelas:
This changelist should come with a testcase or at least test log updates that shows that one of the existing test has a meaningful change in its UHDM output.

RRozak:
I didn't know about it. I added the testcase.

alaindargelas:
Code formatting is incorrect.
I'll fix it post merge. Now that we have the code formatting in place, remember to run it :-)

PR #1406, opened 2021-06-22
test for Ariane elab bug

alaindargelas:


alaindargelas:
Surelog does not match Verilator's elaboration tree, this test to track the diff (no_check generate block is not correct)

PR #1507, opened 2021-07-16
Treat slStringConst as vpiRefObj in compileVariable function

RRozak:


alaindargelas:
Invalid fix for a wrong problem, see comments in Issue.

PR #1529, opened 2021-07-23
Prepare parse listener generator for out-of-tree build.

hzeller:
Having separate input and output directories allows to choose
to read inputs from one diretory and output to another.


hzeller:
The windows-msvc build fails, but this seems to be unrelated to this change (and also happens in head)

PR #1545, opened 2021-07-27
Add helper files for YCM

pieter3d:
The softlink to build/compile_commands.json and the .ycm_extra_config file will enable users with the YCM plugin for vim to have automatic code completion and source browsing (go to declaration etc).

The contents of the .ycm_extra_config are pretty standard fare from the YCM project


tgorochowik:
It seems that after this PR was merged the github pages build started to break, e.g.: https://github.com/chipsalliance/sv-tests/pull/1677/commits/c36fbde42e8137b306424d13c8e9c530ab7715ca

It worked before this PR.

The issue seems to be that the link points to a file that does not exist yet - could we for example generate this file dynamically? That is generate this symlink together with the original file?

@pieter3d would this work for you?

pieter3d:
Yeah generating the link at build time would work just as well.

PR #1546, opened 2021-07-27
Unmark errors as reported right when created

pieter3d:
Currently errors are marked as reported right when the report text is created. This means that when the compiler runs with muteStdOut enabled, the errors are never displayed and never programmatically obtainable, since `ErrorContainer::createErrorMessage` exits right away if the error is marked as reported.

This change removes the error mark when the report text is created. It should be safe, since the only time the private `ErrorContainer::createReport_` methods are called, the calling functions already set the `m_reported` to true anyway. Without this change, that was redundant.


alaindargelas:
This change is incorrect, see diffs in the regression

pieter3d:
Woops - fixed that

PR #1551, opened 2021-07-27
Smallish warning fixes

hzeller:
* Variables that stay uninitialized in some branches.
 * Missing 'break' in switches.
 * missing [[fallthrough]] in switches.
 * Fix a comparison meant to be an assignment.
 * Remove unused variables.


hzeller:
Please confirm that the [`[[fallthrough]]`](https://github.com/chipsalliance/Surelog/pull/1551/files#diff-844a11b8cd04a7c5b8ae86342587f09c72ae151b4b5d791a2fc1c0ff90b60000R1423) was indeed meant to be as such, or if this was a bug and it actually meant to be a `break`

PR #1552, opened 2021-07-27
Fix parameters with correct type to compileTypespec.

hzeller:


hzeller:
Calls to the [`compileTypespec()`](https://github.com/chipsalliance/Surelog/blob/master/src/DesignCompile/CompileHelper.h#L211-L216) method seem to be off - an empty string at the end was interpreted as boolean `true` by accident, so this first version of this pull request just replaces that with what the compiler sees, however, this is certainly due to some shift in added parameters at some point.

The default value of that parameter `isVariable` would be false. Is that the intended value ? If so, then I can just remove the `true` at the end.

hzeller:
(marked as draft, as it is not meant to be submitted until the intended value has been determined)

alaindargelas (commented):
It should be false instead of true

hzeller:
Alright, removed the parameter entirely, so that the default value for that parameter (`false`) is used.

PR #1561, opened 2021-07-29
Avoid including uhdm.h in headers to reduce compile time.

hzeller:
The uhdm.h header includes every other class, which are
mostly not needed in the surelog headers. Typically, we
only need forward-declared containers, so include that
instead.

### DIFF
--- a/src/Design/ModuleDefinition.h
+++ b/src/Design/ModuleDefinition.h
@@ -34,7 +34,7 @@
 #include "Design/ModPort.h"
 #include "Design/Signal.h"
 #include "Design/ValuedComponentI.h"
-#include "uhdm.h"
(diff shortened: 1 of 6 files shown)


alaindargelas:
Can you verify that you enjoy the same code completion (IDE proposes all the UHDM methods) in all the files in DesignCompile/*.c* in IDEs (VSCODE)?
If not, I can't accept this PR.

hzeller:
Don't know, I have never used an IDE with code completion.
But the places where you need the classes, the uhdm.h header is included.

alaindargelas:
I don't see much of a compile time improvement looking at the history of the Actions compile

hzeller:
Still it is a bad idea to include headers in places where they are not needed. IWYU, but not more.

PR #1578, opened 2021-07-31
Add TODO for object leackage in preprocess harness.

hzeller:
While at it, have preprocess file take a string_view


hzeller:
The test-failure looks like independent of this change.

PR #1587, opened 2021-08-01
Reduce leaks in CompileExpression_test

hzeller:
There are more systematic leaks going on inside the libraries linked,
so this test is not leak-free yet, but at least make sure to
deallocate the ojects created in the test.


hzeller:
The CI failure seems to be unrelated (it is also in head), as we only modify a unit-test here.

PR #1591, opened 2021-08-01
Document leak of MacroInfo

hzeller:
Ownership is a bit fuzzy, so for now just document to address
later.


hzeller:
No functional change which would change outcome of CI, just adding a comment.

PR #1616, opened 2021-08-07
Add more specific dependencies for the nix build.

hzeller:
gperftools (for tcmalloc) and zlib are also requested
by cmake.

diff and time are used in the regression tests.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -22,6 +22,10 @@ pkgs.mkShell {
       antlr4
       pkg-config
       libuuid
+      diffutils
+      time
+      gperftools
+      zlib
     ];
 
 }


hzeller:
CC: @sjkelly

sjkelly:
Good catch. The test packages are small, but if you ever need larger and optional dependencies in the shell you can add arguments (right now `pkg` is the only argument. For example:

```
{pkgs ? .... ,
 tests ? false}:

pkgs.mkShell {
  buildInputs = with pkgs;
    [ 
       ...
    ]
    ++ (lib.optional tests my_test_dep1)
    ++ (lib.optional tests my_test_dep2);

```

And call with `nix-shell --arg tests true` to enable the specific dependencies for the `tests` flag.

PR #1625, opened 2021-08-08
Removing leading backslash from array instances

pieter3d:
Rationale: The IEEE1800 doc says that the leading backslash is not part of the identifier, just a way to write the identifier in source code.

Note: I had to hack `regression.tcl` since for some reason it was not stripping my local run directory properly.


alaindargelas:
We should remove the trailing space too?

pieter3d:
Yes, pushed.

PR #1636, opened 2021-08-10
Add macOS build target

rkapuscik:


rkapuscik:
Closing for now to avoid triggering multiple builds on every push (the one on the fork is enough).

PR #1650, opened 2021-08-14
Restore fast antlr and tcmalloc for Linux

alaindargelas:


tests/FuncArgDirection/FuncArgDirection.log
```diff
@@ -871,7 +871,6 @@ design: (work@dut)
     \_io_decl: (in), parent:work@dut.aes_rev_order_bit
       |vpiName:in
       |vpiDirection:1
-      |vpiSize:66
```
hzeller, line 874: Are these logs meant to be changed by this PR ?

alaindargelas: Yes, those logs where not updated for some reason, that is a good change..... but....
I'm going to can this pull request altogether after working really hard on it to figure out that my optimized anltr fork is actually used, Surelog points to alainmarcel/antl4 which contains my optimizations. I forgot what was done by @hs-apotell on this topic.
The only thing that was not working is that the libtcmalloc was not linked in because of a cmake conditional directive not working.

Issue #1526, opened 2021-07-22
External CMake C++ project depending on Surelog 


pieter3d:
Hello! It's not clear to me what the best way is to have an external C++ project pull in Surelog.
What I've done so far:
- set an environment variable SURELOG_ROOT where surelog is checked out and built.
- Add the following to my CMakeLists.txt: 
```cmake
find_library(SURELOG_LIBRARIES NAMES surelog  PATHS $ENV{SURELOG_ROOT}/build/lib)
```
- Depend on this with my target:
```cmake
target_link_libraries(my_executable PUBLIC ${SURELOG_LIBRARIES})
```

This results in a lot of linker errors (e.g. flatbuffers, UHDM) for all of Surelog's third_party libraries. I started to add those one by one, but it started getting messy, especially with Antlr since it's not built within Surelog's `build/` directory (somehow ends up in `dist/`). 

Checking out Surelog as a submodule to my project results in collisions with my project's googletest (age-old cmake problem), and fixing this also gets messy. 

What is the recommended way to do this?

hs-apotell:
Try CMake install.
```
cmake --install <folder path>
```

This will create a distribution which pulls in all the necessary headers and libraries under one root. You should be able to map your system environment variable to this new location.

Pitfalls - This is not where you build Surelog and so you can't have a continuous build and run cycle. You will have to build *and install* each time you happen to modify Surelog.

alaindargelas:
I'm guessing the preferred method would be to have Surelog as a subsystem and make it install it's lib and headers in a standard fashion? Who can help?

hs-apotell:
> ... Surelog as a subsystem ....

Do you mean [exporting](https://cmake.org/cmake/help/latest/guide/tutorial/Adding%20Export%20Configuration.html) and [packaging](https://cmake.org/cmake/help/latest/guide/tutorial/Packaging%20Debug%20and%20Release.html)?

alaindargelas:
I'm no expert, but it should be like Surelog uses cappnp or antlr

pieter3d:
I suppose also Surelog should be exporting a CMake module file in some way, that way other projects can use Surelog libraries with `find_package`

pieter3d:
Here's a workaround for anyone wanting to include Surelog's installed libraries and includes with their CMake project:

```cmake
find_package(Threads REQUIRED)
find_library(SURELOG_LIBRARIES NAMES surelog
  PATH_SUFFIXES surelog) 
[...]

PR #1660, opened 2021-08-15
Issue #1526: Use Surelog as an external package with CMake

hs-apotell:
Issue #1526 
Use Surelog as an external package with CMake

* Updated both UHDM & Surelog to expose relevant/public include paths
  so they get correctly included in the generated output config file.
* Fix the issue where uhdm.h was included twice in the installation.
* Following file in any project would pull in surelog as a dependency
  including setting up the include paths.

  find_package(surelog PATHS \<path to where the lib was installed\>)

  Similarly, appending the module path for CMake would also work.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -207,9 +207,13 @@ jobs:
         where ninja && ninja --version
 
         make release
+        if %errorlevel% neq 0 exit /b %errorlevel%
         make install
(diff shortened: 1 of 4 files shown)


hs-apotell:
@alaindargelas NOTE: This change is dependent on another one in UHDM repo https://github.com/chipsalliance/UHDM/pull/472 Make sure to update UHDM before you merge this pull request.

PR #1662, opened 2021-08-16
Add a unit-test for `ifdef

hzeller:


hzeller:
Interesting is, that this test showed that the lines in the not taken branch are eliminated, which could create issues with line-counts later on the preprocessed output, unless otherwise managed.

alaindargelas:
The line count is indeed managed in the code and accounts for that.

PR #1670, opened 2021-08-16
Issue #1526: Using Surelog as external project dependency

hs-apotell:
Issue #1526: Using Surelog as external project dependency

* Updated both UHDM & Surelog to expose relevant/public include paths
  so they get correctly included in the generated output config file.

* Fix the issue where uhdm.h was included twice in the installation.

* Following line in any project's CMakeLists.txt would pull in surelog
  as a dependency including setting up the include paths.

  find_package(surelog PATHS <path to where the lib was installed>)

* Similarly, appending the module path for CMake would also work.

* Updated README.md with instructions on how to include Surelog as
  external module with CMake.

* Windows test builds had been failing with compile errors but the error
  wasn't being propagated up correctly, and so the builds were always
  green on dashboard. Fixed the underlying compiler issue and also
  updated the workflow file so the errors are propagated correctly.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -207,9 +207,13 @@ jobs:
         where ninja && ninja --version
 
         make release
+        if %errorlevel% neq 0 exit /b %errorlevel%
         make install
+        if %errorlevel% neq 0 exit /b %errorlevel%
         make test_install
+        if %errorlevel% neq 0 exit /b %errorlevel%
         make test/unittest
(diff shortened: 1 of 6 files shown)


hs-apotell:
@alaindargelas Make sure to merge the UHDM change https://github.com/chipsalliance/UHDM/pull/473 before merging this one. This PR also has the same change as in the mentioned UHDM PR.

alaindargelas:
Interrupted jobs and rerunning with UHDM PR merged

PR #1679, opened 2021-08-19
Add a macOS build target

rkapuscik:
This PR adds a CI target to build Surelog on macOS 10. The flow was also locally tested to work on macOS 11 (M1).

I've also opened https://github.com/google/flatbuffers/pull/6802 to not rely on a patch for FlatBuffers.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -222,6 +222,51 @@ jobs:
         name: surelog-windows
         path: ${{ github.workspace }}/install
 
+  macos:
+    runs-on: macos-latest
+
+    steps:
+    - uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+
+    - uses: actions/checkout@v2
+      with:
+        submodules: recursive
+        fetch-depth: 0
+
+    - name: Configure shell
+      run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -222,6 +222,53 @@ jobs:
...
+        python-version: 3.8
+
+    - uses: actions/checkout@v2
+      with:
+        submodules: recursive
+        fetch-depth: 0
+
+    - name: Configure shell
+      run: |
+        echo 'CC=gcc-9' >> $GITHUB_ENV
```
hzeller, line 240: With the [recent change](https://github.com/chipsalliance/Surelog/commit/1bc87b8250347b6ad595055ba022a73a1d47aa8b) in making a regular expression work in libstdc++, would this maybe even work with the native clang on MacOS ?
It might be a good way to test something with a different enough set-up to be more confident that things are robust.

(and if it does not work with clang++, we should collect issues we see in a bug, so that they can be addressed)

hzeller: (might need to rebase the antmicro fork to see the latest change and test there)

rkapuscik: Added a separate build in #1685 - with a small patch it seems to be working.

PR #1685, opened 2021-08-20
Add MacOS clang build

rkapuscik:
This PR renames the previously added macOS build to `macos-gcc` and adds a separate clang build target.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -260,14 +260,14 @@ jobs:
         make release
         sudo make install
 
-    - name: Regression tests
-      run: |
-        make regression
-
     - name: Unit tests
       run: |
         make test_install
         make test/unittest
+
+    - name: Regression tests
+      run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -260,15 +260,56 @@ jobs:
...
+        python-version: 3.8
+
+    - uses: actions/checkout@v2
+      with:
+        submodules: recursive
+        fetch-depth: 0
+
+    - name: Configure shell
+      run: |
+        echo 'PREFIX=/usr/local' >> $GITHUB_ENV
```
hzeller, line 287: For testing install, I think it would be good to have `PREFIX` something other than `/usr/local`, maybe some `/tmp/some-directory`, because that will help us find issues with the install process.
(also for the other installation place inthe linux install).

(But this can be a separate pull request if there are issues changing it now)

PR #1698, opened 2021-08-23
Run test, install and regression in separate matrix builds.

hzeller:
* In Linux CI: run tests, regression and install as separate parallel runs.
  * Use ccache on Linux and Mac which speeds up repeated builds in prticular of low-change source dramatically
  * Cancel previous actions when new pushes come in while actions are still running.
  * Get the number of parallel processes on Mac using the system local means; that allows to run with more `CPU_CORES` than 1

Overall, this makes the CI run much faster for Linux and Mac.
 TODO: something equivalent to ccache on the Windows builds (paging @hs-apotell :) )

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,12 +6,29 @@ on:
 
 jobs:
   linux:
+    name: Check · ${{ matrix.mode }}
+
     runs-on: ubuntu-latest
     defaults:
       run:
         shell: bash
+    strategy:
+      fail-fast: false
+      matrix:
+        mode:
+        - test
+        - regression
+        - install
+    env:
+      MODE: ${{ matrix.mode }}
 
     steps:
+
+    - name: Cancel previous
+      uses: styfle/cancel-workflow-action@0.8.0
+      with:
+        access_token: ${{ github.token }}
+
     - name: Install dependencies
       run: |
         sudo apt-get update -qq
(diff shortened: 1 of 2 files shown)


alaindargelas:
I don't see any logs for the actual build. What happens if the build fails? Did you forget to register a step?

hzeller:
What do you mean, the logs are at the usual place in the action. E.g. if you click in the Check Test thing on Test, the build log of that unfolds, etc.

hzeller:
Ah I see, you want the build broken out separately. I put them together with the actual thing being tested, so that we can choose different compilation modes, e.g. install and regression uses 'release', while tes/tunittest-d uses 'debug'. This keeps things nicer together.

If there is an issue, we will see easily ,as the error will be at the end of the log message anyway, so no difference to now.

hs-apotell:
Shouldn't the branch name be part of the cache key? Repository is built for multiple different branches and if the process caches and restores only the last build then it is likely not very optimal. Also, the caching should run regardless of build success or failure (the compilation was done either way, so if the compilation succeeded then the output should be good).

hzeller:
The cache is meant to be shared between instances. The cache only caches exactly the same input file -> output file association, so most files are typically good.

The ccache will indeed store the cache independent of success or failure of the rest.

hs-apotell:
The default cache size is 500 MB as per the documentation. The output build folder on Windows is over 3GB. I don't have a linux box handy to check the size but I am guessing it is larger than 500 MB. The cache size might need some tweaking to be effective.

PR #1764, opened 2021-08-30
Reduce visual clutter compile messages in CI build.

hzeller:
Don't show the percentage and each file that is compiled,
just output warnings and errors.


hzeller:
it is unclear why the install check fails, will investigate that first.

Issue #1727, opened 2021-08-27
Surelog parsing seems pretty slow


pieter3d:
Example, parsing the Swerv-EH1 core on an i5-9600K 3.7GHz takes about 18 seconds.
I tried generating the valgrind callgrind data, but I gave up waiting after a few minutes. Profiling a smaller design looks like below in kcachegrind. Seems like it's mostly all in the antlr parser. Anything that can be done here?

![image](https://user-images.githubusercontent.com/598246/131088286-3bd328d4-c28e-4fef-b431-177dd392c53e.png)

hs-apotell:
Unless someone else is volunteering for it, I am taking charge of swapping the default RTTI for something custom and "hopefully" faster. I have already started the work and should have something to showcase in next day or two. Tracking #1718

hs-apotell:
@pieter3d Could you please revisit the performance metrics and compare them against your previous ones. Thanks!

pieter3d:
Tried parsing a design with the latest vs from 5 days ago: 19.4s -> 18.5s (about 1 second faster).

alaindargelas:
@hs-apotell I'm sure @pieter3d  is looking for something more drastic like 2x to 5x faster. Let's keep this open.

pieter3d:
For reference, I was parsing and elaborating the swerv-eh1 core.

alaindargelas:
Let's keep this issue and closing #273

siliconvoodoo:
More data point on the improvements?

hzeller:
The biggest improvements can certainly be achieved by reducing ambiguities in the grammar (one example e.g. sketched in https://github.com/chipsalliance/Surelog/pull/2756 by @QuantamHD ); this is as ambiguities results in exponential parsing costs as there might be a lot of back-tracking needed.
Right now, the grammar is written so that it is easy to understand and close to the spec, but that is of course not necessarily optimal and there is a lot of room for improvement reducing ambiguities and look-ahead for someone who has some experience in Antlr grammars.
Depending on the particular code to be parsed, this can probably improve some parsing examples by 10x-100x. Someone with Antlr expertise, please help out :)

[...]

Issue #1718, opened 2021-08-26
Disabling RTTI


hs-apotell:
RTTI is enabled by default for both gcc & cl compiler. Also, dynamic_cast is being used across the codebase. Use of RTTI at library forces enabling RTTI at the program level which, ideally, we would like to disable it for number of reasons including performance and size.

There are options that can be used to restrict RTTI information to only the classes that need it. Here's an [implementation](https://github.com/DummySphere/cpp-rtti). I am endorsing this implementation but just presenting one such option.

Any thoughts on implementing custom RTTI rather than forcing it to be enabled at program level?

P.S. We use CopperSpice (a fork of Qt) for our UI and the output binary sizes grow by a factor of 2.8 times by enabling RTTI at program level. This is not an ideal solution but UDHM/Surelog is the only library we use that forces enabling RTTI.

alaindargelas:
@hs-apotell any update/path to recoup the 24% runtime increase? If not we have to revert back to stock antlr (Not my fork) with RTTI in the main Surelog as the slowdown is not acceptable on Linux.
You can point to your RTTI-less Antlr implementation in your fork. For the antlr generated code in Surelog and the surelog code itself, I think we can use your RTTI replacement as the dynamic casts are not in the millions per run like in Antlr. 
But first we need to recoup the 24% loss. Please update on your findings.

hs-apotell:
Sorry, juggling multiple priorities. I haven't had chance to evaluate the performance of the solution just yet. As a stop gap solution, can we just point Surelog to use the master branch of Antlr instead of the modified Surelog branch. That would mean using pristine Antlr without the bitfield modifications. I will try get the RTTI version working in next few days. This is critical change for us as well which has to get done but at the immediate moment I need to clear my backlog so I can dedicate some exclusive time on this.

If you are comfortable with this stop gap solution, either drop a comment here so I can create the PR with necessary change or update the .gitmodules yourself.

alaindargelas:
Yes please do create a PR with default main antlr

alaindargelas:
OK, after merging the stock Antlr here are the runtime measurements:

Stock Antlr runs the regression in **1980s**
Antlr with RTTI removed using @hs-apotell's technique runs in **1660s**
My version with the bitfields runs in **1306s.**

I'm adding back the bitfield version under third_party/antlr_fast and make the Surelog cmake point to it meanwhile @hs-apotell experiments with the stock antlr/RTTI-less antlr.
[...]

PR #1769, opened 2021-09-01
Issue #1718, #1727 Switch to using a custom implementation of RTTI (antlr specific for now)

hs-apotell:
Issue #1718, #1727
Surelog parsing is pretty slow

Resolution
Replacing use of standard RTTI i.e. use of dynamic_cast
with a custom implementation that avoid having to enable
RTTI across the entire application. This changes replaces
it for antlr only and future change would do the same
for whole of Surelog.

NOTE: The compile provided RTTI is not yet disabled.


hs-apotell:
@alaindargelas Need a bit of help from you. Getting some test failures deep in the parsing code. Basically, all the tests related to ExprEval are failing. I scanned thru' my changes in Antlr a number of times and I didn't find any issues there.

P.S. The Linux builds are failing at the moment and I am working on getting them to pass. The compiler on Linux is stricter than the one on Windows.

alaindargelas:
How do I see the changes you make in Antlr?
You probably need to get it to compile for Linux before I can try it.

hs-apotell:
> How do I see the changes you make in Antlr?

I have opened 3 PRs on https://github.com/alainmarcel/antlr4. The changes are also merged into this PR.

> You probably need to get it to compile for Linux before I can try it.

Yes I am on it. Having a rough time getting antlr built on Linux. Some not so informative error message. Will post back once I have it building.

hs-apotell:
@alaindargelas https://github.com/chipsalliance/Surelog/actions/runs/1194464058
The compile itself is succeeding but now the post-build step for OVM/UVM is failing. Here's the snippet from the log -
```
[SNT:PA0207] 1800.2-2017-1.0/src/base/uvm_globals.svh:335:6: Syntax error: extraneous input '(' expecting {Pound_Pound_delay, Pound_delay, ATSTAR, AT_PARENS_STAR, ';', 'type', 'local', 'super', '{', '->', 'if', 'foreach', 'byte', 'shortint', 'int', 'longint', 'integer', 'time', 'void', '$', '++', '--', DOLLAR_UNIT, '(*', 'assert', 'assume', 'cover', 'expect', 'disable', '##', '#', 'begin', 'end', 'case', 'for', 'assign', 'deassign', 'force', 'release', 'fork', 'repeat', '@', 'return', 'break', 'continue', 'wait', 'wait_order', 'unique', 'unique0', 'priority', 'casez', 'casex', 'randcase', 'forever', 'while', 'do', 'restrict', ''', 'randsequence', 'this', DOLLAR_ROOT, 'randomize', 'sample', '->>', Escaped_identifier, Simple_identifier, SURELOG_MACRO_NOT_DEFINED},
      ("UVM/INIT/MULTI", "Non-recoverable race during uvm_init")

      ^-- ./slpp_all/work/src_4515915290766961624\uvm_pkg.sv:9818:6:.

[INF:CM0023] Creating log file ./slpp_all/surelog.log.

[SNT:PA0207] 1800.2-2017-1.0/src/base/uvm_globals.svh:335:6: Syntax error: extraneous input '(' expecting {Pound_Pound_delay, Pound_delay, ATSTAR, AT_PARENS_STAR, ';', 'type', 'local', 'super', '{', '->', 'if', 'foreach', 'byte', 'shortint', 'int', 'longint', 'integer', 'time', 'void', '$', '++', '--', DOLLAR_UNIT, '(*', 'assert', 'assume', 'cover', 'expect', 'disable', '##', '#', 'begin', 'end', 'case', 'for', 'assign', 'deassign', 'force', 'release', 'fork', 'repeat', '@', 'return', 'break', 'continue', 'wait', 'wait_order', 'unique', 'unique0', 'priority', 'casez', 'casex', 'randcase', 'forever', 'while', 'do', 'restrict', ''', 'randsequence', 'this', DOLLAR_ROOT, 'randomize', 'sample', '->>', Escaped_identifier, Simple_identifier, SURELOG_MACRO_NOT_DEFINED},
      ("UVM/INIT/MULTI", "Non-recoverable race during uvm_init")

      ^-- ./slpp_all/work/src_4515915290766961624\uvm_pkg.sv:9818:6:.
```

Not sure if this is the reason for failure or not. Appreciate some pointers to assist in moving ahead on this.

alaindargelas:
Since you updated to the master there are way too many changes to look at.
Have you tried the updated from master before you made your changes? Was it working?
Did you change the path to the java library to match the one used in the master? The java code generator needs to match the C++ runtime library.
Also, can you disable the pre-compiled package build in the cmake and only try a small verilog example?

hs-apotell:
> Since you updated to the master there are way too many changes to look at.
Have you tried the updated from master before you made your changes? Was it working?

No I haven't yet. Will give that a shot. I was just hoping the issue might be something obvious and don't have to go thru' the process of reverting the changes and all.

> Did you change the path to the java library to match the one used in the master? The java code generator needs to match the C++ runtime library.

Yes, both are rebuilt binaries.

> Also, can you disable the pre-compiled package build in the cmake and only try a small verilog example?

I already tried that. As I mentioned in my earlier message, some of the unit tests are failing.

hs-apotell:
> Since you updated to the master there are way too many changes to look at.
Have you tried the updated from master before you made your changes? Was it working?

Just verified that the build passes (including all tests) with just upgrading the antlr repository to 4.8.3. Something in my changes is causing a failure. Will investigate further my making my changes in smaller steps to see if I can narrow down the issue.

Thomasb81:
> I already tried that. As I mentioned in my earlier message, some of the unit tests are failing.

Is latest commit fix remaining functional issues ?

Have you some figures of performance improvement, just to get a rough idea (memory footprint, runtime) ?

hs-apotell:
Yes all the functional issues are resolved and all tests are succeeding. As for numbers, I haven’t done a thorough diagnosis but preliminary runs show roughly about 10% improvements. This task isn’t fully complete yet because UHDM and Surelog are still using compiler provided RTTI. Need follow up changes to switch those projects (and turn off compiler based RTTI) as well to realize full benefits of this change.
________________________________

alaindargelas:
The 10% improvements came from a couple of grammar optimization I did yesterday.
On the other hand, after merging all of these RTTI changes, things got much slower today:

On an 8 Cores i9 processor at 4.5Ghz:

Yesterday with just the grammar tuning the whole regression ran in:
TOTAL ELAPSED 1306s    

With RTTI changes from today:
TOTAL ELAPSED 1624s    

That is **24% slower** than my custom bitfield in antlr!

hs-apotell:
This implementation (with some tweaks) comes from number of products we have shipped over the years and is battle tested. It would be very concerning if the implementation is a bottleneck. Let me run some local tests today.
________________________________

alaindargelas:
The version I had yesterday as a baseline was the new antlr version you merged. 
That run was very comparable with my old fork and the bit field, antlr must have sped up things since I forked.

After merging your RTTI replacement, things got 24% slower.

Maybe it's a Linux vs Windows thing? Your implement is faster on Windows and slower on Linux?

hs-apotell:
The issue with the bitfield solution is that it doesn't work with virtual multiple inheritance.

The implementation I have added to the codebase has shipped with multiple games and across many platforms including Windows, Linux, Xbox, Playstation and some Nintendo platforms too. If I am not wrong a similar implementation is also part of Unreal code base. I don't believe the implementation is flawed. There is something more to it and I will investigate further.

mithro:
I do want to mention that things in the C++ world continue to evolve as C++ compilers continue to get better (especially after LLVM became a viable competitor to GCC). There are plenty of examples where it **used** to be better to do things yourself that are no longer the case (there is a reason that browsers run with tip-of-tree LLVM these days).  I'm unsure if RTTI falls into this case, but it certainly wouldn't surprise me if something that was significantly faster 5 years ago is now unable to beat the default options.

hzeller:
@hs-apotell please make sure that your changes are abstracted enough to work with stock Antlr4.
As long as you modify the local antlr4 copy to generate a different code that behaves the same on the outside, that is fine (I am packaging Surelog in environments that can _only_ work with stock Antlr, not with a hacked local one; so code generation and compilation should work with both).

As I said before, we should aim for building a fast runtime for Antlr and upstream that (e.g. don't use RTTI stock or fake, limit creating lots of temorary objects etc.), _then_ use it here as it will be much more relevant for everyone using Antlr.

hs-apotell:
> C++ world continue to evolve as C++ compilers continue to get better

Here's a [comparison](https://kahncode.com/2019/09/24/c-tricks-fast-rtti-and-dynamic-cast/) of RTTI across different compilers. It is certainly still possible to implement a custom version of it and gain substantial benefits.

> please make sure that your changes are abstracted enough to work with stock Antlr4.

I will include a compile time flag to turn this change on/off. From the outside, the library isn't really changing. The classes have a few (two to be precise) new virtual functions which are used to up/down cast and global functions used in place of `dynamic_cast`. Everything should be fairly easy to replace wrap it in a compile time macro.

> As I said before, we should aim for building a fast runtime for Antlr and upstream that (e.g. remove all the code-smells: don't use RTTI stock or fake, limit creating lots of temorary objects, don't use exceptions etc.), then use it here as it will be much more relevant for everyone using Antlr.

That would need some dedicated resource to pull this off. Unfortunately, I don't think we have anyone on our team with parser/compiler knowledge good enough to pull this off.

hs-apotell:
@alaindargelas I have an improved version of the RTTI ready for review. I see that you have abandoned alaindargelas/antlr4 repository and pulled in a copy of it again within Surelog, What's the plan here? Is there a new forked chipsalliance/antlr4 repository that I can publish the change for review?

A quick update on progress ---

In summary of the change - I piggybacked on the bitfield change that we already had in the repository (and currently have in antlr_fast). The bitfield logic is used for single inheritance tree where is works well and for multiple-inheritance and virtual inheritance, a more elaborate implementation was required. On Windows, this has gained about 7% improvement on regression times. The big drawback with the bitfield logic is someone/something has to know the entire inheritance tree to avoid stomping (in current implementation the base class has an enumeration listing all subclasses). This solution wouldn't scale well for in-production or still-in-development code.

Unfortunately, the RTTI still cannot be disabled entirely because of the dependency on exceptions. They both seem to go hand-in-hand. It came as learning for me too that exceptions are dependent on RTTI and typed-exceptions would crash if RTTI was disabled. So, in summary, we need more work to completely disable RTTI in antlr.

alaindargelas:
You can use alaindargelas/antlr4 to develop your RTTI changes, essentially I left that for you to take over. The local copy in antlr_fast is there to stay untouched until we have another solution from you under alaindargelas/antlr4. 
There is no chipsalliance/antlr4 at this point. We can pursue that later.

PR #1786, opened 2021-09-03
Upgrade antlr to v4.8.3

hs-apotell:


hzeller:
The UI in github is a littly confusing as it doesn't really tell what the antlr4 submodule is pointing to.

I _suspect_ this should only work after
https://github.com/alainmarcel/antlr4/pull/1 is in and then the Surelog branch is rebased to it, but it seems the CI is happy anyway, but I am a bit confused as the pull request is not yet in ?

Since the github UI doesn't, can you tell us the antlr branch we submodule and point to ?

CMakeLists.txt
```diff
@@ -132,7 +132,7 @@ add_custom_command(
 # Java
 find_package(Java REQUIRED COMPONENTS Runtime)
 set(ANTLR_JAR_LOCATION
-    ${PROJECT_SOURCE_DIR}/third_party/antlr4/antlr-4.9.1-complete.jar)
+    ${PROJECT_SOURCE_DIR}/third_party/antlr4/antlr4-4.9.3-complete.jar)
```
hzeller, line 135: If the files that are submitted to the repository in `src/parser/*` (for people who don't want java to be installed in the build process), these generated files should probably also be updated

(though can be a separate pull request)

(we could also should think about if we actually want to ship the pre-generated parser files or build them as part of the process always, that avoids freshness issues (e.g. weird timestamps of the checked out *.g4 vs. *.cpp files), and one can argue if someone compiles Surelog, they are also happy to install everything needed with it)

hzeller:
The title of the pull request says 4.8.3, but it looks like it means 4.9.3. You can change the commit message with

```
git commit --amend
git push -f
```

CMakeLists.txt
hs-apotell: I personally believe that developers should be willing to run the entire process rather than making auto-generated files part of the source deploy. It tends to hide the dependencies for not much gain. Moreover, Java setup is trivial regardless of platform and its only JRE not even the JDK requirement. At some point, we should really consider moving these auto-generated files out of the source folder and into the build directory.

Java is sort of a necessity for developers these days. If anything, Tcl is more of a pain on windows.

hs-apotell:
> alainmarcel/antlr4#1

This change was pulling from remote repository antlr/antlr4 into the alaindargelas/antlr4 master branch. Note, Surelog doesn't use this branch.

> alainmarcel/antlr4#2

This change was merge from alaindargelas/antlr4 master to alaindargelas/antlr4 Surelog branch which is actually used by Surelog

> #1786

This change was to update the Surelog submodule

> The title of the pull request says 4.8.3, but it looks like it means 4.9.3. You can change the commit message with

Sorry the change got merged before I could act.

CMakeLists.txt
hzeller: I agree, so we should consider removing the files in `parser/`. @alaindargelas WDYT ?

hs-apotell: Another reason to either remove or move these file into the build folder is that Windows support multiple configurations and lot of our developers tend to run both debug & release builds in parallel which isn't possible with files being in source folders.

Similar problem exist in UHDM project as well. Both headers and src folders are in project's home directory. Those need to be moved into the build directory as well.

hzeller: Fully agree. The source folders should be considered readonly for the purpose of the build; all generated artifacts should end up in `build/` somewhere. Including code generated by the tcl scripts.

alaindargelas:
Sure, please do.

    On Friday, September 3, 2021, 10:23:05 PM PDT, Henner Zeller ***@***.***> wrote:  
 
 


@hzeller commented on this pull request.

In CMakeLists.txt:
> @@ -132,7 +132,7 @@ add_custom_command(
 # Java
 find_package(Java REQUIRED COMPONENTS Runtime)
 set(ANTLR_JAR_LOCATION
-    ${PROJECT_SOURCE_DIR}/third_party/antlr4/antlr-4.9.1-complete.jar)
+    ${PROJECT_SOURCE_DIR}/third_party/antlr4/antlr4-4.9.3-complete.jar)

I agree, so we should consider removing the files in parser/. @alaindargelas WDYT ?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

PR #1800, opened 2021-09-05
hier_path binding

alaindargelas:


mysoreanoop:
Hi Alain,
I believe we are having an issue with structs in SV code. What we see is that if there's a struct member reference within the control expression of an vpiIfStmt, it doesn't resolve into full instance scope unlike with all other register/wire references. Here's an example if it helps:
```
if(some_wire) begin //we get the full instance scope for some_wire like work@top.(parent_inst.)*local_inst.some_wire
if(some_struct.member) begin //doesn't resolve like above; instead work@some_struct.member
```

Even within a uhdm-dump, I'm unable to resolve struct members' full scope. Any idea what could be wrong? One non-invasive way we're planning to tackle this is by saving the struct instances in every model and referencing that when we encounter structs as part of control expressions. Does that sound like a decent way to accomplish this? How would you do it?

Thanks a tonne in advance!
Anoop

alaindargelas:
Please add complete example that compiles and show the problem in UHDM. The example should be the smallest possible but still show the problem. Add a visual text pointer (<----) to the incorrect UHDM output too.

mysoreanoop:
Updated, thank you

alaindargelas:
Also, I should have mentioned, this is a new issue, please do not append to closed issue, open a new issue.

mysoreanoop:
> Also, general practice, this is a new issue, please do not append to closed issue, next time open a new issue. UHDM does not do bit blasting of the object if that is what you are after.
Very sorry, thought this might be related. 
> 
> If I rework your example to have both cases:
> 
> * The hier_path object is used in a control statement,
> * The hier_path object is used as a variable
> 
> You can see from the dump below, that you get the same information in both cases, nothing special about the case statement, a hier_path object with 2 ref_obj members, and for each members their binding to the struct and the field in the struct.
> 
> What else would you need to be able to re-create the full hierarchical path?
> 
Oh, okay.. I understand now. I missed that there are 2 ref_objs. This is sufficient for determining the full paths. Mighty thanks!

> Note from your dump, it looks like you don't have the -elabuhdm on which binds the hier_path elems to the variable declaration

PR #1815, opened 2021-09-08
Issue #1718: Fallback to pristine antlr

hs-apotell:
Issue #1718: Fallback to pristine antlr

Introduced RTTI solution seems to be slower than the standard
language implementation using dynamic_cast. There is no reason
for this to be the case and needs investigation. For the time
being, as a stop gap solution, falling back to pristine Antlr
to not have all developers pay the penalty.


hs-apotell:
All checks passed on my [forked repo](https://github.com/hs-apotell/Surelog/actions/runs/1214753699) but one failed here. Looks like machine ran out of disk space. Kicked off another build just to confirm.

PR #1858, opened 2021-09-14
Put generated outputs into build/ folder.

hzeller:
This will generate the code generated by the parser generator
and the vobject-types tcl script in a folder in the build/
directory.
Remove the previously checked-in pre-generated parser files. They
are now always created at build time.


hzeller:
CC @hs-apotell you might like this one.

hs-apotell:
That's a welcome change.

PR #1898, opened 2021-09-20
Add another workflow to allow running regressions on-demand

hs-apotell:
Supports running Windows and Linux regressions on demand.


.github/workflows/regression.yml
```diff
@@ -0,0 +1,123 @@
+name: 'regression'
+
+on:
```
hzeller, line 3: Can you add a comment that this is meant for on-demand triggering ?

hs-apotell: Renamed the file and the workflow itself.

PR #1932, opened 2021-09-26
Memory Optimization Pass

hs-apotell:
Memory Optimization Pass

Use std::string_view in place of std::string in SymbolFactory.
Updated all code generation scripts to use std::string_view.

Updated build script to build googletest with C++17 as well so
we are using the same compiler settings across the entire project.
This is also required to bring in support for std::string_view.

NOTE: This change includes chipsalliance/UHDM#517

PR #1960, opened 2021-09-29
Make SymbolTable return stable std::string& on query.

hzeller:
There are a few places that take strings coming from the SymbolTable
and then modifying the very SymbolTable that then can make these
string references render invalid due to reallocations within the
table.

Change that, so that we have realloc-safe strings by storing pointers
in the underlying vector instead of string objects. These are the
stable backing of all the string references handed out so we can
guarantee that these are valid as long as the SymbolTable is valid.

As a nice side-effect, we don't have to store strings, just string_views
as keys in the symbol -> ID map.

Due the unfortunate fact that we have copy constructor and we don't
want to re-create all the strings, use a shared_ptr to reference
count these. On the plus side, copies are now cheaper.

More expensive became getSymbols() that expects a std::vector<std::string>
to be returned due to some requirement by flatbuffers. Ideally this
can be changed to a std::vector<std::string_view> at some point.

### DIFF
--- a/src/Cache/Cache.cpp
+++ b/src/Cache/Cache.cpp
@@ -162,9 +162,8 @@ Cache::cacheErrors(flatbuffers::FlatBufferBuilder& builder,
   }
 
   /* Cache all the symbols */
-  for (auto itr = symbols->getSymbols().begin();
-       itr != symbols->getSymbols().end(); itr++) {
-    canonicalSymbols.registerSymbol(*itr);
+  for (const auto& s : symbols->getSymbols()) {
+    canonicalSymbols.registerSymbol(s);
   }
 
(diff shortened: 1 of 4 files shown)


hs-apotell:
@hzeller Please refer to pull request #1932.

hzeller:
Ah, nice, looks like you also had a stab at adding string_views @hs-apotell .

Here I am solving a problem that made @alaindargelas change the `collectFiles()` parameters from `const std::string&` to `std::string` by value, as the function itself added symbols, that invalidated of course string references on re-alloc of the internal vector (Looks like #1932 will have the same problem).

I agree, we should return string-views, but I didn't want to do this in this change, as there are too many changes needed for that. Looks like you're on it.

PR #1962, opened 2021-09-29
Add a clang-tidy run to the CI

hzeller:
Using a baseline of useful clang-tidy warnings, but with
some disabled that would be too noisy right now.

For now, this is just generating an FYI output as
a starting point to get these adressed. Once the baseline
is clean, we can change this to result in an CI error if
new warnings show up in a pull request.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,50 @@
+---
+# readability-make-member-function-const is great, but it also suggests that
+#    in cases where we return a non-const pointer.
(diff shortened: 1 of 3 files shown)


alaindargelas:
Can we make so it does not fail? This is going to be bugging me if suddenly all PR are fail and we have to go see which action failed.

hzeller:
No, it will not fail at this point.

Issue #1979, opened 2021-09-30
Use std::filesystem instead of manhandling paths as strings


hs-apotell:
We are noticing quite a few issues emerging from use of strings to manage paths. Here are few instances,

* https://github.com/chipsalliance/Surelog/blob/094f1ad629c34d38c0c578f1a8c913eb8d7c1d82/src/SourceCompile/CompileSourceFile.cpp#L94
* https://github.com/chipsalliance/Surelog/blob/094f1ad629c34d38c0c578f1a8c913eb8d7c1d82/src/SourceCompile/CompileSourceFile.cpp#L126

Also, Windows limit of 260 characters applies when using strings are arguments to classes like std::ofstream but it doesn't when using std::fs::path. We are running into the path limit on Windows with certain tests (most recently with BlackParrot) and using std::fs seems to resolve it. Unfortunately, the documented solution to fix the path limit doesn't seem to work for Surelog.

hzeller:
+1 for using fs::path to assemble file names.

alaindargelas:
The BlackParrot test path is not a Surelog specific issue, it is a BlackParrot test structure issue.

alaindargelas:
Yup, please update the minimum gcc, send a PR.

hs-apotell:
For sake of documentation - Found another reason justifying this change. Surelog calling external processes using `system` command and using backslashes on Msys2 causes failure. Specific test case failing as of today is `DiffSimpleIncludeAndMacros` where the input to the `diff` command isn't being found because the input output directory on the command-line was using backslashes.

alaindargelas:
Fixed by #2253

PR #2025, opened 2021-10-07
Issue #1979,  chipsalliance/UHDM#510

hs-apotell:
Issue #1979,  chipsalliance/UHDM#510

* Upgrade CI to use gcc9 instead of gcc7
* Also, updated the regression_on_demand to run the regression
  using both pygen and tclgen.
* Moved the setting to use specific version of C++ before adding
  subprojects to ensure all dependent projects use the same C++
  version.
* Windows specific - MSVC-cl has a bug where the value of
  __cplusplus is wrong unless a specific compile flag is included.
  gtest requires the value to be correct to enable inclusion of
  string_view specific matchers. Forcing the compile flag from
  Surelog to avoid having to clone/fork gtest. More information on
  this bug can be found here -
  https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-160
  https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,7 +34,7 @@ jobs:
       run: |
         sudo apt-get update -qq
         sudo apt install -y \
-          g++-7 \
+          g++-9 \
           tclsh \
           default-jre \
           cmake \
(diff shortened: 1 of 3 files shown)


hs-apotell:
I will make it part of regular regression once the changes settle down. I am still working on some of the parts of transition to python. I don't want CI builds to fail and increase pressure on myself to fix them immediately. I am juggling too many competing priorities so it's not always feasible to attend to fires. It will be part of regular CI in due time.

hzeller:
Sounds good. 
How can we trigger it now in the context of this PR ?

PR #2067, opened 2021-10-14
Memory Optimization Pass

hs-apotell:
Memory Optimization Pass

Use std::string_view in place of std::string in SymbolTable.
Also, optimize the storage by ensuring we don't store multiple
copies of same string.


hzeller:
This should be broken in down in various changes to be reviewable each addressing one of the things

  * Start with the symbol table, just change the internal implementation if that is the core of the improvement. If there are string_views first to be returned, convert them to `std::string` when returning to keep the interface. So the firt change should only include the SymbolTable, its header and additions to the tests. Describe how the change will improve memory usage.
  *  There seems to be changes adding mutexes implying some multi-threaded changes. Do that in a separate change.
  * There are other changes throughout that should also be addressed separately (e.g. avoiding string copies and instead assigning things to `const std::string&`. This is entirely separate aspect and should be broken into its own change.

hs-apotell:
> Start with the symbol table, just change the internal implementation if that is the core of the improvement. If there are string_views first to be returned, convert them to std::string when returning to keep the interface. So the firt change should only include the SymbolTable, its header and additions to the tests. Describe how the change will improve memory usage.

Current implementation of `SymbolFactory` returns a reference to a `std::string` and calling site is usually holding this reference (correct/expected behavior) instead of cloning the returned string. Returning a `std::string` if I am storing a `std::string_view` internally would be reference to local variable. Plus, `std::string_view` to `std::string` is not an implicit conversion and so I can't just change the API's return type. Bottom-line, I don't think this transition can happen without touching every site where the `SymbolFactory` is queried. How do propose changing `SymbolFactory` alone would work?

> There seems to be changes adding mutexes implying some multi-threaded changes. Do that in a separate change.

This is implementing the internal storage - multiple read, single write operation - since the storage itself is shared across multiple instances of `SymbolFactory`. This is the core of the `SymbolFactory` change.

I will address each of the concern raised individually.

hs-apotell:
> Can you quantify the savings impact ?
> 
> Even if we have 1M symbols and would have an overhead of one allocated string per symbol instead of pointing into a congiguous area, we're still only talking about maybe 12 MiB overhad (or whatever `sizeof(std::string)` is) vs. this approach of having a few contigjuous areas.

The savings are multi-fold because the storage is shared across `SymbolFactory` instances. At least, that's the intent here. The cache backing is a fall-out from what I noticed when debugging `black-parrot` regression test. The `SymbolFactory` instance gets cloned nearly 600+ times so each string in the original `SymbolFactory` is cloned nearly that many times. Shared storage avoid that. The contiguous block itself is to prevent fragmentation when working with larger designs (and avoid putting all the burden on global allocator like tcmalloc).

> I think we should address the string copies all over first by actually making the rest of the code base ready: have all parameters that currently accept `const std::string &` (or even `std::string`) accept string_view.
> 
> Then, it will be much easier to also have functions to _return_ string_view as they can be processed by other places. At that stage, a change of SymbolTable to return string-views will be hopefully not requring much effort.

May be coming from the other end was better, or may be not. Either way most if not all of these files would be affected. I took the approach of attack the root and see what fails and fix that. May be other approach, attacking a class, a file, would have been granular but I wouldn't have been able to measure the overall impact of the change - which I am able to do now and I see some negatives that needs addressing. There are pros and cons to either approach - I just happen to choose one.

hs-apotell:
[Attached](https://github.com/chipsalliance/Surelog/files/7360963/Perf.csv) is the complete report of comparison of two Surelog/master builds with two Regression builds that include this change. And, below is the quick summary (selected all tests which report over 1GB memory use).


* Negatives in below table are savings in both time and memory
* PyGen generated code reports faster runtime (relative to Tcl generated code) when compared to current master runtime.
* Memory savings are not very high even for large tests. In some cases (like Google & YosysTests) the consumption is considerably higher. No memory leaks found in this change.
* Interesting edits in this change are primarily in SymbolTable implementation.
* Implementation includes no mutexes

```
Surelog/master builds: 2442 & 2452
Regression builds: 17 & 18

                            |                 |                 |    |                      |                       |                       |                         |    |    Surelog: 2442     |    Surelog: 2442      |    |     Surelog: 2452    |     Surelog: 2452     |    |    Surelog: 2442     |     Surelog: 2442     |    |    Surelog: 2452     |     Surelog: 2452     |
                            |  Surelog: 2442  |  Surelog: 2452  |    | PyGen Regression: 17 | TclGen Reggresion: 17 | PyGen Reggresion: 18  | TclGen Reggression: 18  |    |        vs.           |           vs.         |    |          vs.         |          vs.          |    |        vs.           |          vs.          |    |         vs.          |          vs.          |
                            |                 |                 |    |                      |                       |                       |                         |    | PyGen Regression: 17 | TclGen Regression: 17 |    | PyGen Regression: 17 | TclGen Regression: 17 |    | PyGen Regression: 18 | TclGen Regression: 18 |    | PyGen Regression: 18 | TclGen Regression: 18 |
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
TESTNAME                    | TIME  | MEM(Mb) | TIME  | MEM(Mb) |    |   TIME   |  MEM(Mb)  |   TIME   |   MEM(Mb)  |   TIME    |  MEM(Mb)  |    TIME    |  MEM(Mb)   |    |    TIME  |  MEM(Mb)  |    TIME   |  MEM(Mb)  |    |    TIME  | MEM(Mb)   |    TIME   |  MEM(Mb)  |    |    TIME  | MEM(Mb)   |    TIME   |  MEM(Mb)  |    |   TIME   |  MEM(Mb)  |    TIME   |  MEM(Mb)  |
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
AmiqEth                     |    16 |    1018 |    21 |    1014 |    |       15 |      1013 |       19 |       1008 |        15 |      1006 |         20 |      1008  |    |       -1 |        -5 |         3 |       -10 |    |       -6 |        -1 |        -2 |        -6 |    |       -1 |       -12 |         4 |       -10 |    |       -6 |        -8 |         -1|        -6 |
Ariane                      |    50 |    1434 |    62 |    1432 |    |       49 |      1426 |       60 |       1426 |        50 |      1425 |         60 |      1428  |    |       -1 |        -8 |        10 |        -8 |    |      -13 |        -6 |        -2 |        -6 |    |        0 |        -9 |        10 |        -6 |    |      -12 |        -7 |         -2|        -4 |
AzadiRTL                    |    43 |    1066 |    55 |    1069 |    |       43 |      1065 |       49 |       1066 |        45 |      1064 |         54 |      1066  |    |        0 |        -1 |         6 |         0 |    |      -12 |        -4 |        -6 |        -3 |    |        2 |        -2 |        11 |         0 |    |      -10 |        -5 |         -1|        -3 |
BasicOh                     |       |    1066 |       |    1069 |    |          |      1065 |          |       1066 |           |      1064 |            |      1066  |    |        0 |        -1 |         0 |         0 |    |        0 |        -4 |         0 |        -3 |    |        0 |        -2 |         0 |         0 |    |        0 |        -5 |          0|        -3 |
BuildUVMPkg                 |   217 |    1608 |   239 |    1614 |    |      216 |      1608 |      216 |       1610 |       218 |      1603 |        238 |      1610  |    |       -1 |         0 |        -1 |         2 |    |      -23 |        -6 |       -23 |        -4 |    |        1 |        -5 |        21 |         2 |    |      -21 |       -11 |         -1|        -4 |
Compl1001                   |   114 |    2564 |   134 |    2571 |    |      113 |      2565 |      128 |       2566 |       116 |      2565 |        132 |      2565  |    |       -1 |         1 |        14 |         2 |    |      -21 |        -6 |        -6 |        -5 |    |        2 |         1 |        18 |         1 |    |      -18 |        -6 |         -2|        -6 |
CoresSweRV                  |    35 |    1565 |    43 |    1566 |    |       32 |      1553 |       38 |       1551 |        33 |      1554 |         41 |      1552  |    |       -3 |       -12 |         3 |       -14 |    |      -11 |       -13 |        -5 |       -15 |    |       -2 |       -11 |         6 |       -13 |    |      -10 |       -12 |         -2|       -14 |
Earlgrey                    |    29 |    1377 |    37 |    1378 |    |       28 |      1378 |       32 |       1374 |        29 |      1372 |         35 |      1379  |    |       -1 |         1 |         3 |        -3 |    |       -9 |         0 |        -5 |        -4 |    |        0 |        -5 |         6 |         2 |    |       -8 |        -6 |         -2|         1 |
Earlgrey_0_1                |    41 |    1948 |    51 |    1945 |    |       38 |      1944 |       44 |       1947 |        39 |      1941 |         48 |      1943  |    |       -3 |        -4 |         3 |        -1 |    |      -13 |        -1 |        -7 |         2 |    |       -2 |        -7 |         7 |        -5 |    |      -12 |        -4 |         -3|        -2 |
Earlgrey_Verilator_0_1      |    43 |    2031 |    53 |    2031 |    |       39 |      2023 |       46 |       2025 |        41 |      2023 |         50 |      2021  |    |       -4 |        -8 |         3 |        -6 |    |      -14 |        -8 |        -7 |        -6 |    |       -2 |        -8 |         7 |       -10 |    |      -12 |        -8 |         -3|       -10 |
Earlgrey_Verilator_01_05_21 |    66 |    1096 |    84 |    1096 |    |       61 |      1136 |       72 |       1135 |        64 |      1138 |         78 |      1137  |    |       -5 |        40 |         6 |        39 |    |      -23 |        40 |       -12 |        39 |    |       -2 |        42 |        12 |        41 |    |      -20 |        42 |         -6|        41 |
Google                      |    93 |    1554 |   118 |    1558 |    |       95 |      1908 |      107 |       1866 |        96 |      1923 |        112 |      1820  |    |        2 |       354 |        14 |       312 |    |      -23 |       350 |       -11 |       308 |    |        3 |       369 |        19 |       266 |    |      -22 |       365 |         -6|       262 |
Ibex                        |    30 |    1188 |    37 |    1194 |    |       28 |      1188 |       35 |       1187 |        29 |      1186 |         35 |      1189  |    |       -2 |         0 |         5 |        -1 |    |       -9 |        -6 |        -2 |        -7 |    |       -1 |        -2 |         5 |         1 |    |       -8 |        -8 |         -2|        -5 |
IncompTitan                 |    50 |    1990 |    59 |    1988 |    |       45 |      2005 |       52 |       2006 |        46 |      2003 |         57 |      2007  |    |       -5 |        15 |         2 |        16 |    |      -14 |        17 |        -7 |        18 |    |       -4 |        13 |         7 |        17 |    |      -13 |        15 |         -2|        19 |
Opentitan                   |    39 |    1642 |    46 |    1643 |    |       35 |      1634 |       40 |       1642 |        36 |      1640 |         44 |      1639  |    |       -4 |        -8 |         1 |         0 |    |      -11 |        -9 |        -6 |        -1 |    |       -3 |        -2 |         5 |        -3 |    |      -10 |        -3 |         -2|        -4 |
Rsd                         |    69 |    3762 |    81 |    3763 |    |       58 |      3753 |       69 |       3756 |        61 |      3753 |         76 |      3752  |    |      -11 |        -9 |         0 |        -6 |    |      -23 |       -10 |       -12 |        -7 |    |       -8 |        -9 |         7 |       -10 |    |      -20 |       -10 |         -5|       -11 |
Unisim                      |   123 |    1166 |   145 |    1167 |    |      116 |      1260 |      136 |       1247 |       119 |      1250 |        143 |      1261  |    |       -7 |        94 |        13 |        81 |    |      -29 |        93 |        -9 |        80 |    |       -4 |        84 |        20 |        95 |    |      -26 |        83 |         -2|        94 |
YosysOpenSparc              |    61 |    4092 |    75 |    4065 |    |       61 |      4097 |       73 |       4074 |        62 |      4194 |         74 |      4106  |    |        0 |         5 |        12 |       -18 |    |      -14 |        32 |        -2 |         9 |    |        1 |       102 |        13 |        14 |    |      -13 |       129 |         -1|        41 |
YosysSmallBoom              |    32 |    3263 |    37 |    3270 |    |       29 |      3240 |       33 |       3297 |        29 |      3254 |         35 |      3243  |    |       -3 |       -23 |         1 |        34 |    |       -8 |       -30 |        -4 |        27 |    |       -3 |        -9 |         3 |       -20 |    |       -8 |       -16 |         -2|       -27 |
YosysTests                  |    50 |     825 |    59 |     828 |    |       47 |      1196 |       55 |       1329 |        48 |      1169 |         58 |      1322  |    |       -3 |       371 |         5 |       504 |    |      -12 |       368 |        -4 |       501 |    |       -2 |       344 |         8 |       497 |    |      -11 |       341 |         -1|       494 |
```

alaindargelas:
@hs-apotell I really don't understand the purpose of this changelist.
Some tests show very minor memory savings, but some tests show huge increase of memory. Does not look like a good trade at all.

alaindargelas:
My suspicion is that you are keeping handles to memory in the symbol table so it does not show as memory leaks. It shows as memory used at exit. That memory is accumulating for each design parsed in batch mode.

alaindargelas:
Has the memory blowout been eliminated?

hs-apotell:
Nope. Just re-based it. I spent another day on it yesterday but no go. I don't see what/how the memory could increase so much. There are absolutely no leaks and I have even compared the life-time of SymbolTable from master and this branch. They are identical. Nothing obvious stands out.

I am also seeing a similar pattern with another change I have in the works. It seems like allocating bigger blocks of memory significantly bumps up the peak memory usage. Can't justify it but that's what it seems.

alaindargelas:
Maybe using mmap would help
https://stackoverflow.com/questions/41256789/is-memory-malloc-through-mmap-returned-to-os-immediately-after-calling-free

Large chunks might not be released back to the OS with new/malloc.

hs-apotell:
I will give that a shot at least to confirm the hunch. Will figure out a more portable solution (mmap calls are not portable across platforms) if that happens to be the problem.

hs-apotell:
Abandoned!!!!

PR #2215, opened 2021-11-19
Improve python regression script parity

hs-apotell:
Improve python regression script parity with Tcl and results across
different platforms.

Few tweaks to bring parity between Tcl and Python regression runs.
These were the issues identified as part of the case study to compare
the results across both implementation. Highlights include -

* Catch all exceptions in thread to avoid Python from error-ing on the
  entire process. Caught exceptions are logged to regression.log file
  so they can individually be diagnosed.
* Msys2 doesn't like windows style directory separators.
* Msys2 also doesn't like Windows specific paths - it actually wants
  paths that are unix-like which forces the use of cygpath.
* Tests involving shell scripts were consistently failing. It needed
  the 'sh' as the first parameter for them to succeed on Msys2.
* Report the longest and largest test in summary at the end of the
  regression.
* Double-quotes (and quotes, in general) needed to be escaped

### DIFF
--- a/.github/workflows/pymain.yml
+++ b/.github/workflows/pymain.yml
@@ -129,20 +129,33 @@ jobs:
         make clean   # make sure we only see installation artifacts
         make test_install
 
+    - name: Prepare build artifacts
+      if: matrix.mode == 'install'
+      run: |
+        mv install surelog-linux-gcc
+        mkdir build
+        tar czvfp build/surelog-linux-gcc.tgz surelog-linux-gcc
+
+    - name: Prepare regression artifacts
+      if: matrix.mode == 'regression' && always()
+      run: |
+        cd build
+        mv regression surelog-linux-gcc-regression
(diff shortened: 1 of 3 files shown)


alaindargelas:
This looks like it is ready for merge?

hs-apotell:
No. Msys2 still has an issue with DecValue test. Attempting to resolve it.

alaindargelas:
macos build is broken with this PR:
make: nproc: Command not found

hs-apotell:
@alaindargelas I am not sure I follow you. The error you mentioned happens on both main and pymain workflows. nproc is not a legal command on macos and so the error is legal and is safely ignored by CMake. The macos failures with main workflow have nothing to do with this PR. They have been happening on current master as well. This PR introduces no changes to main workflow. All changes in this PR are to python regression script and pymain workflow, which did succeed as part of this PR's checks.

https://github.com/chipsalliance/Surelog/runs/4309216787?check_suite_focus=true

PR #2233, opened 2021-11-26
replace non-ascii by space

alaindargelas:


Thomasb81:
Does that mean that source code can contain non ascii character ?
According to annex A.9.2, even comments should restrict to ascii symbol...

alaindargelas:
There is the standard, and what the commercial tools do support, some files contain non ascii character in comments and are widely accepted.
Surelog was too restrictive.

Thomasb81:
I observe and understand that.

From my point of view :
It has value to be able to track/log such discrepancy with the standard  : It remains work : implement standard, analyze issues, find work-around.
Such a list can by itself serve as promotional material as could be api documentation, command line help for the tool...

alaindargelas:
#2243 adds a Note when non-ascii character is found.

PR #2277, opened 2021-12-08
static func

alaindargelas:


alaindargelas:
It's correct default lifetime is static, if you add the automatic keyword before the function, you'll see the automatic property.

alaindargelas:
It's correct default lifetime is static, if you add the automatic keyword before the function, you'll see the automatic property.

PR #2418, opened 2022-01-10
Fixed hardcoded output folder for tests

hs-apotell:
Antlr - Don't create the library in the dist folder. This creates problems for multi-configuration build systems.
Tests - Fixing two tests where the inputs provided had typos.


hs-apotell:
@alaindargelas This merge is going to need new log for 'YosysOldI2c` test. The input had a typo and causing the test to fail across all platforms.

PR #2422, opened 2022-01-11
Use https instead of git to clone utfcpp

kamilrakoczy:
Github recently removed support for using git protocol unauthenticated: https://github.blog/2021-09-01-improving-git-protocol-security-github/.
This PR fixes clone of utfcpp, example error: https://github.com/hdl/conda-eda/runs/4772215688?check_suite_focus=true#step:3:1646
Signed-off-by: Kamil Rakoczy <krakoczy@antmicro.com>


kamilrakoczy:
sv-tests are also affected by this: https://github.com/chipsalliance/sv-tests/runs/4769677699?check_suite_focus=true#step:16:309

tgorochowik (approved):

PR #2522, opened 2022-02-02
Improve readability: use empty() when checking for the emptiness of c…

hzeller:
…ontainers.

Depending on the container, this can also be more efficient, but mostly this
is to improve readability and express intent better.

https://clang.llvm.org/extra/clang-tidy/checks/readability-container-size-empty.html


src/SourceCompile/AnalyzeFile.cpp
```diff
@@ -125,7 +125,7 @@ std::string AnalyzeFile::setSLlineDirective_(unsigned int lineNb,
     result += std::to_string(1);
     result += "\n";
   }
-  result = "";
+  result = "";  // BUG or intentional ?
```
hzeller, line 128: here I have a question: independent of what happens above, here, the result is cleared out again.
Is this intentional or the leftover from some previous debugging ?

Issue #2575, opened 2022-02-11
doesn't build clean on ubuntu20


mattvenn:
Following the instructions from simview:

Install pre-requisite packages (Ubuntu 20.04 LTS)
sudo apt install build-essential cmake git pkg-config tclsh swig uuid-dev libgoogle-perftools-dev python3 python3-dev default-jre lcov libncurses-dev
Build and install [Surelog](https://github.com/chipsalliance/Surelog)
git clone git@github.com:chipsalliance/Surelog.git
cd Surelog
git submodule update --init --recursive
make
sudo make install

I get this error:

[ 38%] Linking CXX executable capnpc-c++
make[3]: Leaving directory '/tmp/Surelog/build'
[ 38%] Built target capnpc_cpp
make[3]: Entering directory '/tmp/Surelog/build'
Scanning dependencies of target GenerateCode
make[3]: Leaving directory '/tmp/Surelog/build'
make[3]: Entering directory '/tmp/Surelog/build'
[ 38%] Generating generated/src/UHDM.capnp
Traceback (most recent call last):
  File "/tmp/Surelog/third_party/UHDM/scripts/generate.py", line 7, in <module>
    import loader
  File "/tmp/Surelog/third_party/UHDM/scripts/loader.py", line 1, in <module>
    import orderedmultidict
ModuleNotFoundError: No module named 'orderedmultidict'
make[3]: *** [third_party/UHDM/CMakeFiles/GenerateCode.dir/build.make:386: third_party/UHDM/generated/src/UHDM.capnp] Error 1
make[3]: Leaving directory '/tmp/Surelog/build'
make[2]: *** [CMakeFiles/Makefile2:1207: third_party/UHDM/CMakeFiles/GenerateCode.dir/all] Error 2
make[2]: *** Waiting for unfinished jobs....
[ 38%] Generating generated/src/SourceCompile/VObjectTypes.h, generated/src/SourceCompile/VObjectTypes.cpp, generated/src/API/VObjectTypes_py.h, generated/src/SourceCompile/SV3_1aTreeShapeListener.h, generated/src/SourceCompile/SV3_1aPpTreeShapeListener.h
---------- VObjectTypes, Listener generation
make[3]: Leaving directory '/tmp/Surelog/build'
[ 38%] Built target GenerateParser
make[2]: Leaving directory '/tmp/Surelog/build'
make[1]: *** [Makefile:163: all] Error 2
make[1]: Leaving directory '/tmp/Surelog/build'
[...]

PR #2577, opened 2022-02-11
CI/INSTALL.md: Add missing Python3 packages needed to build.

hzeller:
Also use the same instructions in the CI to make sure
things will work as advertised.

Issues: #2575


hzeller:
(only failure in the windows CI, but that is an independent problem)

PR #2613, opened 2022-02-17
synth contruct checker

alaindargelas:


alaindargelas:
@rkapuscik, you can add this code to your Yosys plugin,
Set the bool to false, so it does not create errors on the stdout.
The non-synthesizable nodes are in the set.
Let me know what else you want to mark as non-synthesizable.

https://github.com/chipsalliance/Surelog/blob/d7f4699f9405656977857b71447b58f2fe58cefc/src/DesignCompile/UhdmWriter.cpp#L2756

PR #2631, opened 2022-02-19
Consistency in command line argument parsing

hs-apotell:
Consistency in command line argument parsing

Unquote and strip command line arguments consistently. The args coming
from a file weren't being stripped.

Also, don't use strstr for arg comparison. When looking for arguments
like -D, -I, -P, and alike, the arg has to match at position 0, not
at any position within the string.

After considering all options, if no match is found, ignore the arg
if it starts with a hyphen (anything starting with hyphen should be
considered control and not a value)


hs-apotell:
@alaindargelas Log file for Ariane test case needs to be updated. The number of warnings has changes and I have verified them to be legitimate.

PR #2691, opened 2022-03-02
Fixing location issues

hs-apotell:
Fixing location issues

* Fixed missing location information
* Fixed use of incorrect NodeId for retrieving location


hs-apotell:
Regression failure is expected. Logs will need to be updated.

PR #2699, opened 2022-03-03
Don't edit directories chosen by flags.

hzeller:
Concatenating directoris or making them absolute restricts
the usefulness of the script as it prevents relative
output paths that don't share a prefix with the working
directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -192,7 +192,7 @@ add_custom_command(
 
   WORKING_DIRECTORY ${PROJECT_SOURCE_DIR}
   COMMAND echo "---------- VObjectTypes, Listener generation"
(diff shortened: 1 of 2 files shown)


hs-apotell:
With this change what will be your working directory? Workspace directory or output directory? It becomes ambiguous. The assumption is that all relative paths are w.r.y workspace directory and otherwise are absolute. It also helps with keeping the number of arguments needed to change directory to a minimum.

Recommend thinking thru' all the different scenarios especially multi configuration builds like windows where 'build' is not the default output directory.

hzeller:
Exactly that is why I did the change: in my case the build directory is also in a different place and the realpath of the workspace directory is not a prefix of the output (as this is a symbolic link in my case).

I'd actually rename the workspace directory more like 'input directory' or 'source base', and the other 'generate base' or something (similar to the original tcl script)

PR #2754, opened 2022-03-14
Update UhdmChecker.cpp

little-blue:
when I build UHDM-yosys,I always meet error in this file,I think it forget `#include <iomanip>`, after add this will not meet error when building ,here's the error 
```
yosys-uhdm-plugin-integration/Surelog/src/DesignCompile/UhdmChecker.cpp:249:42: error: ‘setprecision’ is not a member of ‘std’
，
yosys-uhdm-plugin-integration/Surelog/src/DesignCompile/UhdmChecker.cpp:315:67: error: ‘setw’ is not a member of ‘std’
```

### DIFF
--- a/src/DesignCompile/UhdmChecker.cpp
+++ b/src/DesignCompile/UhdmChecker.cpp
@@ -40,9 +40,10 @@
 #include <uhdm/vpi_visitor.h>
 
 #include <fstream>
+#include <iomanip>
 #include <sstream>
 #include <stack>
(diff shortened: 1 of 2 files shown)


alaindargelas:
I'll make the change.
That won't work :-)
[Update and rename src/DesignCompile/UhdmChecker.cpp to 源代码/设计编译/UhdmC…]

alaindargelas:
Fixed by #2771

PR #2796, opened 2022-03-29
Fixing more location information

hs-apotell:
Fixing more location information. Also, fixed dependency issue in CMakeLists.txt and a few const-ness.


alaindargelas:
Please fix the conflict.

alaindargelas:
@hs-apotell, please attach a complete regression.log file.

hs-apotell:
You can download all logs from the CI build. All the artifacts are here -
https://github.com/chipsalliance/Surelog/actions/runs/2057030041

alaindargelas:
There is no regression.log file that contains the diff in those artifact. I'd have to reconstruct the diff.
It's not usable to do a diff, sorry, please send a complete diff in a single file, the "regression.log" file.
Thanks

alaindargelas:
Better yet, submit all the log updates part of the PR:

cd build

Your Python script must have the same function.

tests/HierPathSelect/HierPathSelect.log
```diff
@@ -154,13 +159,15 @@ design: (work@dut2)
       \_ref_obj: (read_buf), parent:read_buf.q
         |vpiName:read_buf
       |vpiActual:
-      \_bit_select: (work@dut2.read_buf.q.q), line:8:17, endln:8:18, parent:read_buf.q
```
alaindargelas, line 157: Huge bug here!

alaindargelas (commented):
I'll review some more, but there is already a big bug introduced by this PR, see comment.
It does way too may things at the same time! The title of the PR is misleading. It does way more.

tests/HierPathSelect/HierPathSelect.log
hs-apotell: Isn't what current one also wrong? Shouldn't `work@dut2.read_buf.q.q` be `work@dut2.read_buf.q`? The variable `q` is `logic` so `q.q` doesn't make any logical sense.

I am gonna revert this specific change and let you deal with it. But there are some issues with parenting and actual_group for `bit_select` & `part_select`. For both the parent should have been a `ref_obj` but it wasn't in this specific case which is what I was attempting to fix.

alaindargelas: The .q.q is just a printing bug in the vpiFullName when printing, if you traverse the data model, you can find all the nodes required and to the right thing.
With your change, the nodes are gone, that is a real bug.

Is there any way you can unroll all the "other changes" you deem harmless from this PR and only do what it says in the title: Fixing more location information
Hunting for bugs you might have introduced in 400+ log files in not what I had in mind for this week-end....

hs-apotell: I don't expect you to be spending anytime looking through bugs. That's not your problem. However, please do review the code changes.

I self reviewed my change once again I didn't find anything that wasn't location information specific or _attempt_ to fix bit_select/part_select parenting & actual_group. I have reverted my bad attempt will file bugs against those that I have found.

I have re-based and removed any changes that don't belong with this PR. Feel free to nitpick if you find anything that's not location information specific. Will follow up with an update to logs as well once the build finishes.

alaindargelas:
I still need to see the log updates along with the code change.
Please resubmit the log file changes , even if they are 100s.

PR #2805, opened 2022-04-01
Fix some mechanical readability and performance fixes.

hzeller:
* a few for loops converted to range loop
 * Use some std::string_view or const std::string&
   instead of std::string in parameters.
 * Don't do unnecesary string = "" initialization.
 * Update the limited clang tidy rules that are
   currently running in the CI.


hzeller:
The windows regression is failing due to some path forward/backward slashes, but this looks like a pre-existing condition.

src/Design/FileContent.cpp
```diff
@@ -630,8 +630,8 @@ bool FileContent::diffTree(NodeId root, const FileContent* oFc, NodeId oroot,
     }
     std::string symb1;
     std::string symb2;
-    if (current1.m_name) symb1 = Name(id1);
-    if (current2.m_name) symb2 = oFc->Name(id2);
+    if (current1.m_name) symb1 = std::to_string(Name(id1));
```
hzeller, line 633: In this place, the symbol-ID that was coming back was interpreted as number and so assigned a single char to the `symb1`. Unclear what was meant, but at least I am translating it to the symbol-ID now.

PR #2835, opened 2022-04-10
added vpiInsideOp op for case stmt

alaindargelas:


alaindargelas:
@rkapuscik @kbieganski, please take not to support this new operation

PR #2839, opened 2022-04-11
Introducing RoundTrip - application to validate location info

hs-apotell:
Introducing RoundTrip - application to validate location info

RoundTrip is an application to use the location information from
generated model to regenerate the sources - making it a round trip. The
generated sources are compared against the original to generate statistics.
The application can be run in two independent modes - surelog & uhdm. The
former, Surelog mode includes parsing and the later from binary. RoundTrip
is also part of regression and statistics are reported with every run.
However, the differences (or rather change in differences) wouldn't
report a failure.

The application can also serve as a initial model for implementing a linter.

P.S. This is still a work in progress. A number of open issues where
the required information is missing or in other cases wrong.


alaindargelas:
If I understand, you are doing a pretty printing of UHDM to re-create Verilog, then you have a test harness around it.
The pretty printing code (90% of this code) should be moved to UHDM/utils/PrettyPrinter.h .cpp, it should not be in Surelog.
That way it is usable as a post UHDM application like UHDM/template/UhdmLinter  UHDM/template/Elaborator....
The test harness can stay in Surelog.
Please move the pretty printer code to UHDM, add a section in the UHDM/README and a unit test that exercise it.

hs-apotell:
> pretty printing of UHDM to re-create Verilog

Not exactly since the generated code is not necessarily legal Verilog. But the generated code can be compared against the original (with exceptions) to generate a decent statistics to judge validate location information. For example, semicolons and round-brackets in expressions are unaccounted for in the generated output and the difference is ignored when comparing against the original.

And, by no means the generated code is _pretty_. Assumptions about location of operators, equals and commas doesn't make it a good reading.

> they can rewrite the resulting Verilog making it a powerful framework for code instrumentation or auto-fix kind of application

This can serve as a initial base for something along those lines but this is far from being useful for stated purpose. The sole intention with this implementation is to ensure that location information is available and correct in the model hierarchy and future development can maintain the same.

PR #2846, opened 2022-04-12
Use heterogeneous lookup on map containers to allow string_view lookup.

hzeller:
Prepare some containers to be used with std::string or std::string_view
to look up.

### DIFF
--- a/include/Surelog/Common/Containers.h
+++ b/include/Surelog/Common/Containers.h
@@ -3,12 +3,20 @@
 #pragma once
 
 #include <cstdint>
-#include <functional>
 #include <map>
 #include <string>
+#include <string_view>
 #include <vector>
 
 namespace SURELOG {
+
+struct StringViewCompare {
+  using is_transparent = void;
+
+  constexpr bool operator()(std::string_view a, std::string_view b) const {
+    return a < b;
+  }
+};
 
 class ClassDefinition;
 class ModuleDefinition;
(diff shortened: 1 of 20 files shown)


hs-apotell:
@hzeller I tried this change earlier and as it turns out it has significant performance impact, at least on Windows. The PR was finally dropped because of the performance impact. You might want to run some tests before introducing this in master.

hs-apotell:
Here's the relevant [PR](https://github.com/chipsalliance/Surelog/pull/2067) that was abandoned.

hzeller:
Can you test this PR on a windows machine ?

I've compared this branch and the at head on Linux, and the runtimes are essentially the same within noise (with even a slightly faster run in the heterogeneous look-up).

It is expected that they are at least equals as essentially just the less-than operator compares to sequences of bytes at a time which should note be different between the implementations.

User time for 3 runs Regression test (`TOTAL USER` value in the output of `make regression`)
at head: 1629sec 1628sec 1630sec
with heterogeneous lookup: 1616sec 1615sec 1622sec

The change #2067 does a whole lot more, I suspect if there were performance issues, they might just have been from the other parts ?

hs-apotell:
If I recollect it correctly, the performance was because of the use of heterogeneous comparison operators (I had eliminated everything locally and ran a build on CI as well with just this change). Here's what I [found](https://stackoverflow.com/a/20383485) while researching at the time. The number of temporaries created for comparison can have negative impact. This wasn't available in any of the supported compilers at the time. Can't be sure of the status today.

I can run the test once again for Windows later today. Have you compared the numbers from regression builds from CI for windows target? Do those numbers align the same way as they do for linux-gcc?

You could use the regression_on_demand workflow to kick off parallel runs of both master and your branch for comparison.

hzeller:
There should be no temporaries needed. The `std::less` takes two template parameters of different type is that there are no temporaries needed to do the comparison, and that should work that way since c++14.

I have only compared on my local machine, the CI values are more noisy.

hzeller:
@hs-apotell I've now added an explicit `StringViewCompare` comparator, so that any compiler is guided to do the right thing.

PR #2859, opened 2022-04-15
Don't copy all exisiting entries on every copy constructed SymbolTable.

hzeller:
The SymbolTable's copy constructor is used in the code as snapshotting
the current state, but that is resulting in a huge churn in memory as
a lot has to be copied on each such occasions.

This change makes it cheap as snapshotting starts with empty local
containers and then delegates lookups to parents as needed.
This requires an indirection to the parent if there is one, but this
saves a large amount of memory churn and runtime.

This change saved many gigabytes of memory use and a lot of runtime
while processing a larger project. So very worth-while.

Longer term, all the places that use the SymbolTable's copy constructor
to create a snapshot should be investigated if they can be
solved differently.

Siblings of the same parent are thread safe and can operate concurrently.
This is _not_ thread safe if the parent is modified while the client
does a look-up.

There are simple ways around it, but currently, threaded compiling is
[not done](https://github.com/chipsalliance/Surelog/blob/dc9970a0827a84ce5d7dca60f2264d32503a37f3/src/DesignCompile/CompileDesign.cpp#L282-L288),
so this is not an issue right now. And by the time we can do
multi-threaded compiling again, I hope to have resolved the
SymbolTable copy-need already. So for sake of simplicity, no effort
is put into thread-safety right now.

TESTED
Added comprehensive unit test to make sure that all properties of
the previous symbol table still work and behave exactly the same.
The regreession test passes with no diffs. But faster and using
less memory :)

### DIFF
--- a/include/Surelog/SourceCompile/SymbolTable.h
+++ b/include/Surelog/SourceCompile/SymbolTable.h
@@ -28,7 +28,7 @@
 
 #include <Surelog/Common/SymbolId.h>
 
-#include <memory>
+#include <deque>
 #include <string>
 #include <string_view>
 #include <unordered_map>
@@ -44,7 +44,11 @@ class SymbolTable {
   ~SymbolTable();
 
   // Unfortunately, currently the copy constructor is used in a few places.
-  SymbolTable(const SymbolTable&) = default;
+  // Until that semantic is fixed, make it cheap to refer to higher symbol
+  // tables.
+  // A copy construction essentially works as a 'snapshot' of the copied table.
+  SymbolTable(const SymbolTable& parent)
+      : m_parent(&parent), m_idOffset(parent.m_idCounter + parent.m_idOffset) {}
 
   // Register given "symbol" string as a symbol and return its id.
   // If this is an existing symbol, its ID is returned, otherwise a new one
@@ -57,7 +61,6 @@ class SymbolTable {
 
   // Get symbol string identified by given ID or BadSymbol if it doesn't exist
   // (see #getBadSymbol()).
-  //const std::string& getSymbol(SymbolId id) const;
   const std::string& getSymbol(SymbolId id) const;
 
   // Get a vector of all symbols. As a special property, the SymbolID can be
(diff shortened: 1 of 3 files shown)


hs-apotell:
If I am not wrong windows is threaded compilation. @alaindargelas could you please confirm.

hs-apotell:
As per the comment here in CommandLineParser, this change is likely to cause race conditions or memory corruption.
https://github.com/chipsalliance/Surelog/blob/8bfb36982510d0c2540cd0e411895216b73eaf38/src/CommandLine/CommandLineParser.cpp#L756-L767

QuantamHD:
@hzeller @hs-apotell If this is the case we can implement a snapshot isolation layer in the SymbolTable following this approach.

```
The SymbolTable class has a map from symbols to Ids. 

You can turn this into a vector of maps, and everytime you clone the object you push an empty map onto the end of the vector, and crucially only modify the last map on the vector.

Every child knows what range of the vector it's snapshot is, and when it delegates it loops through all the maps in its range.

Then we have a mutex on the clone operation to make sure that every child has a unique snapshot of the parent symbol table.

This will be a low overhead thread safe operation. If lookups get too slow we can compact the symbol table's between snapshot ids
```

hs-apotell:
There are multiple approaches to avoid copying the data around. One such was implemented in [this PR](https://github.com/chipsalliance/Surelog/pull/2067). This PR had lot of the changes that are being introduced in pieces in last few days. Unfortunately, this specific PR was abandoned because of another memory leak bug in Serializer logic.

The approach used here was to treat the source SymbolTable's existing entries as read-only and use a multi-read, single-write mutex. Since there are limited edits on the source SymbolTable itself, the reads bear a minimum mutex overhead.

QuantamHD:
My approach crucially doesn't require a read mutex at all since the previous map entries are immutable.

hzeller:
If current parsing happens in multi-threading, then we need to do something about it, and @QuantamHD's solution will be the one we would do (as it only requires one mutex lock at copy time).

However, I am not sure if parsing actually happens multithreaded currently, as it is [disabled here](https://github.com/chipsalliance/Surelog/blob/dc9970a0827a84ce5d7dca60f2264d32503a37f3/src/DesignCompile/CompileDesign.cpp#L282-L288).

I think our goal should be to focus on figuring out the places that use the copy right now and if they really have to.

hs-apotell:
> Then we have a mutex on the clone operation to make sure that every child has a unique snapshot of the parent symbol table.

You are contradicting your own statement. :)

Both approaches need mutex just for different operations. Moreover, mutex just around the clone operation is not enough. You have to wrap the read/write operations also with the same mutex because the vector that's holding these maps can get resized/reallocated. If I am correct on this i.e. needing mutex around read/write operations, your solution will likely be more expensive because it will require locking for each parent walking up the hierarchy.

There are compromises to either solution. In my proposed solution, the price is paid for every read/write but the solution is much simpler. In your proposed solution, the price is in compaction which for large designs will become mandatory unless iterating and lookups in 1000 maps somehow becomes cheaper than lookup in two maps wrapped in a mutex.

Regardless, we digress. The point I was trying to make is that this change has already been merged and needs to be followed up with a solution to handle multi-threading.

alaindargelas:
I just tried with 16 threads the parsing stage still works and it does manipulate the symbol tables.
The compilation stage is single threaded so is the elaboration and the preprocessesing unless -fileunit is set

PR #2863, opened 2022-04-16
-nohash saves preprocessing cache full reload

alaindargelas:


hzeller:
I think this patch is not working as intended. After this, attempting to reload files to 'link' them to a UHDM will spew out messages that indicate that preprocessing information is missing `Unknown macro "FOOBAR".`

alaindargelas:
Can you try to remove the if condition here:
https://github.com/chipsalliance/Surelog/blob/1c2459f841f6e6d923b336feacd22ccfb9aea845/src/Cache/PPCache.cpp#L101

The things we are trying to prevent reloading to save memory are really these sections:

https://github.com/chipsalliance/Surelog/blob/1c2459f841f6e6d923b336feacd22ccfb9aea845/src/Cache/PPCache.cpp#L185
https://github.com/chipsalliance/Surelog/blob/1c2459f841f6e6d923b336feacd22ccfb9aea845/src/Cache/PPCache.cpp#L191

hzeller:
If I replace line 101 in PPCache with
```
if (true || !errorsOnly) {
```
this does seem to make it work.

alaindargelas:
Fixed by #2883

PR #2864, opened 2022-04-16
Provide a test for number-parsing of UHDM::constant in getValue().

hzeller:
Writing it, a number of bugs were discovered: the unsigned
binary values (BIN, HEX, OCT) were not actually parsed as
unsigned values. Also no out-of-range detection was done.

Fixed all of these in the course of this change.


hzeller:
I am not sure what is going on with the clang macos test (regression not working), but that is also already upstream so not caused here.

PR #2865, opened 2022-04-16
Start adding a usage message to roundtrip if invoked w/o parameter.

hzeller:
The details need to still be filled in, 'base-dir' and 'log-dir'
are not quite clear what they are.

### DIFF
--- a/scripts/regression.py
+++ b/scripts/regression.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 
 import argparse
 import difflib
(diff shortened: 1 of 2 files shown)


hzeller:
I wanted to play with the roundrip tool, but it is not quite clear to me yet how to use it. So I started a usage message that we can update to be more detailed later.

hs-apotell (approved):
LGTM

alaindargelas:
@hs-apotell how do we update the logs files with the roundtrip section?

hs-apotell:
roundtrip is part of default regression run and updates the log files. To update the logs using regression script 
```
python3 regression.py update
```

PR #2866, opened 2022-04-16
chipsalliance/Surelog#2541: Add include file info model

hs-apotell:
Add include file info model

### DIFF
--- a/src/SourceCompile/SV3_1aPpTreeShapeListener.cpp
+++ b/src/SourceCompile/SV3_1aPpTreeShapeListener.cpp
@@ -206,7 +206,7 @@ void SV3_1aPpTreeShapeListener::enterUnterminated_string(
 void SV3_1aPpTreeShapeListener::enterInclude_directive(
     SV3_1aPpParser::Include_directiveContext *ctx) {
   if (m_inActiveBranch && (!m_inMacroDefinitionParsing)) {
-    std::pair<int, int> lineCol =
+    std::pair<int, int> startLineCol =
         ParseUtils::getLineColumn(m_pp->getTokenStream(), ctx);
     std::pair<int, int> endLineCol =
         ParseUtils::getEndLineColumn(m_pp->getTokenStream(), ctx);
@@ -214,14 +214,20 @@ void SV3_1aPpTreeShapeListener::enterInclude_directive(
     std::string fileName;
     if (ctx->String()) {
       fileName = ctx->String()->getText();
+      startLineCol = ParseUtils::getLineColumn(ctx->String());
+      endLineCol = ParseUtils::getEndLineColumn(ctx->String());
     } else if (ctx->macro_instance()) {
       fileName = m_pp->evaluateMacroInstance(
-          ctx->macro_instance()->getText(), m_pp, lineCol.first,
+          ctx->macro_instance()->getText(), m_pp, startLineCol.first,
           PreprocessFile::SpecialInstructions::CheckLoop,
           PreprocessFile::SpecialInstructions::ComplainUndefinedMacro);
+      startLineCol = ParseUtils::getLineColumn(m_pp->getTokenStream(),
+                                               ctx->macro_instance());
+      endLineCol = ParseUtils::getEndLineColumn(m_pp->getTokenStream(),
+                                                ctx->macro_instance());
     } else {
-      Location loc(m_pp->getFileId(lineCol.first),
-                   m_pp->getLineNb(lineCol.first), 0, 0);
+      Location loc(m_pp->getFileId(startLineCol.first),
+                   m_pp->getLineNb(startLineCol.first), 0, 0);
       logError(ErrorDefinition::PP_INVALID_INCLUDE_FILENAME, loc);
       return;
     }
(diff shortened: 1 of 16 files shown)


hs-apotell:
@alaindargelas Could you please review the approach. Few follow up questions -

* IncludeFileInfo tracks both includes and macros. How to differentiate whether a particular entry belongs to include or macro?
* I am noticing an offset for generated column information. Values are offset-ed by -2. Any clue as to what might be the reason?

Or, instead of differentiating between include and macros, should the model itself have a type property so both include and macro become part of the model hierarchy. Macros will need to be exposed similarly at some point and this might be an opportunity to handle them in the same pass.

src/SourceCompile/SV3_1aPpTreeShapeListener.cpp
```diff
@@ -556,25 +580,28 @@ void SV3_1aPpTreeShapeListener::enterMacroInstanceNoArgs(
...
       }
 
-      IncludeFileInfo info(macroInf->m_line, macroInf->m_file,
-                           m_pp->getSumLineCount() + 1, 0, 0, 0,
-                           IncludeFileInfo::PUSH);
-      m_pp->getSourceFile()->getIncludeFileInfo().push_back(info);
+      m_pp->getSourceFile()->getIncludeFileInfo().emplace_back(
+          macroInf->m_startLine, macroInf->m_file, m_pp->getSumLineCount() + 1,
+          macroInf->m_startColumn,
+          m_pp->getSumLineCount() + 1 +
```
alaindargelas, line 594: I can't do the mental math, but that might be where there is an offset created?

src/SourceCompile/SV3_1aPpTreeShapeListener.cpp
```diff
@@ -621,23 +648,27 @@ void SV3_1aPpTreeShapeListener::enterMacroInstanceNoArgs(
...
       if (nbCRinMacroBody) {
-        IncludeFileInfo infop(line, fileId, m_pp->getSumLineCount() + 1, 0, 0,
-                              0, IncludeFileInfo::POP);
-        infop.m_indexOpening = openingIndex;
-        m_pp->getSourceFile()->getIncludeFileInfo().push_back(infop);
-        if (openingIndex >= 0)
+        m_pp->getSourceFile()->getIncludeFileInfo().emplace_back(
+            line, fileId, m_pp->getSumLineCount() + 1, startLineCol.first,
+            m_pp->getSumLineCount() + 1 +
+                (endLineCol.first - startLineCol.first),
```
alaindargelas, line 662: or here?

alaindargelas (commented):
Don't make any change in the sv_vpi_user.h,
we have another file where we add the UHDM extensions:
templates/uhdm_vpi_user.h
Please move your changes from sv_vpi_user.h to uhdm_vpi_user.h

[include/sv_vpi_user.h](https://github.com/chipsalliance/UHDM/compare/5470f89838bb8f4ad80d6b939d3ae7e33b6230da...5b951a1f6dfc3204d265d22da780abde64a7fea9#diff-a27f14779b7fd9720238b314a93a03229e1589c9654c7870cf8c54edb7c6d8dc)

src/SourceCompile/SV3_1aPpTreeShapeListener.cpp
hs-apotell: It takes the column information directly from value returned from `ParseUtils::getLineColumn` and `ParseUtils::getEndLineColumn`. Only line information is computed and that's coming out correct.

src/SourceCompile/SV3_1aPpTreeShapeListener.cpp
hs-apotell: Same here.

src/SourceCompile/SV3_1aPpTreeShapeListener.cpp
hs-apotell: This seems like a bug in ParseUtils logic. The two overloads for `getLineColumn` and `getEndLineColumn` aren't doing the same thing. Once I match the logic in both implementations, the result comes out of correct.

hs-apotell:
> Please move your changes from sv_vpi_user.h to uhdm_vpi_user.h

Done.

Two questions -
1. `Preprocessor::m_includeFileInfo` collects both file includes and macros. How to differentiate between them? Should I add a enum to disambiguate?
2. In `UhdmWriter`, how do I get the _resolved_ file path? `IncludeFileInfo` only gives the text from the line in source file. How do I resolve this?

alaindargelas:
1. Preprocessor::m_includeFileInfo collects both file includes and macros. How to differentiate between them? Should I add a enum to disambiguate?

Sure.

2. In UhdmWriter, how do I get the resolved file path? IncludeFileInfo only gives the text from the line in source file. How do I resolve this?

I need an example of what you get and what you expect to get.

hs-apotell:
> I need an example of what you get and what you expect to get.

For an example, in PreprocFunc, what I get is "inc.h", what I need is the path to the file that this resolves to.

PR #2870, opened 2022-04-17
Use VObjectType instead of unsigned short for better type safety.

hzeller:
While at it:

  * have getTypeName(VObjectType) return std::string_view
  * Use delegating constructor for VObject to minimize duplication
  * Remove obsolete tcl script to generate parser listener; it
    was confusing which code generator to modify.


alaindargelas:
You should have bumped the version number as the caches are all invalid after this change.
https://github.com/chipsalliance/Surelog/blob/788da2c39d0504ca7d68b53aeec88cbcf269c766/src/CommandLine/CommandLineParser.cpp#L54

PR #2875, opened 2022-04-19
elab and non elab packages

alaindargelas:


hzeller:
Are these changes something that @rkapuscik would need to know details about for the UDHM Yosys plugin ?

alaindargelas:
No, the uhdmTopPackages has not changed and that is what is used by @rkapuscik. The uhdmAllPackages does now preserve expressions symbolically (as requested by @hs-apotell)

PR #2876, opened 2022-04-19
Use Ninja as default; create gcc/clang matrix for tests

hzeller:
Use a matrix with clang and gcc to run test and regression in Linux. Also convert the separate mac-OS targets with the two compilers to a matrix.

Ninja  is typically better than make distributing the
work and the output is focusing on only outputting
information relevant for the developers' attention.


hzeller:
(will be draft for a while while messing with the CI)

PR #2886, opened 2022-04-25
chipsalliance/Surelog#2845: Switch to using pristine ANTLR

hs-apotell:
chipsalliance/Surelog#2845: Switch to using pristine ANTLR

* Deleted local copy of antlr
* Added prebuilt antlr tool to third_party/antlr4_bin
* Updated .gitmodules to point to pristine antlr4/antlr4
* Update CMakeLists.txt to use the updated antlr4
* Updated workflow to use java 11

P.S. third_party/antlr4 is tracking 'dev' branch and not the 'master' as
Surelog is dependent on some of the changes that are not yet released.
Will revisit once the changes in dev are merged into master.


QuantamHD:
Lgtm

PR #2892, opened 2022-04-29
chipsalliance/Surelog#2541: Add include file info model

hs-apotell:
chipsalliance/Surelog#2541: Add include file info model

UHDM::design now provides a list of includes


alaindargelas:
@hs-apotell, Please resolve the conflict and confirm that there is no more regression on the line number

PR #2897, opened 2022-04-30
Regression update

hs-apotell:
Regression update

* Use python regression script in CI
* Right align the numbers in regression report
* Add total-cpu-time in reported metrics

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -102,7 +102,7 @@ jobs:
     - name: Regression
       if: matrix.mode == 'regression'
       run: |
-        make regression
+        make test/regression
 
     - name: Prepare regression artifacts
       if: matrix.mode == 'regression' && always()
@@ -579,7 +579,7 @@ jobs:
 
     - name: Regression tests
       run: |
-        make regression
+        make test/regression
 
     - name: Prepare build artifacts
       run: |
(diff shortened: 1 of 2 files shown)


alaindargelas:
@hs-apotell, there seems to be a problem with one test, maybe it needs to be in the blacklist for Windows?
Can't merge until it is all green.

hs-apotell:
@alaindargelas Something isn't adding up. Windows log is reporting it correct but Linux log is not. The include_file_info is missing for `LowMemPkg` for Linux.

From Windows log -
```
|vpiIncludeFileInfo:
\_include_file_info:  dut.sv:2:10: , endln:2:25
  |vpiIncludedFile:pack_incl.svh
````

which is correct but this snippet is missing in Linux log. 

I don't have a linux setup to test this. @alaindargelas Could do you have a few minutes to dig deeper on Linux front.

hzeller:
Can we remove the wall time measurement/column/max-recording entirely ? It is not very useful (as it essentially only measures how busy the CI machine is), but it adds horizontal space (I wasn't able to display the current output table in the CI on my laptop unless I scaled down the browser font).

Since we're recording the time to get a feel for the performance, CPU-time is the only relevant metric.
(It might be useful to actually display the CPU time with one fractional decimal digit as many tests run for < 1 second.)

PR #2902, opened 2022-05-03
Document cache functions; make more const; fix bug

hzeller:
Start documenting the methods used in the cache. Also

   * Fix a bug: the file_id in the VObject was not properly encoded to the symbol table that was written to disk, but whatever symbol ID it had in the local symbol table (`Cache::cacheVObjects()`).
   * If parameters can be `const`, make them so.
   * Make it impossible to forget delete[]-ing buffer returned by `openFlatBuffers()` by making it a unique_ptr. Simplifies the place where it is used quite a bit.
   * Only emit symbols into the cache that are actually referenced in the elements written to the cache. This reduces file-size and removes other elements that would differ depending on the context (e.g. if the file was compiled together with other files - the additional symbol table entries are not needed and break hermetic build assumptions).
   * Remove timestamp when file was compiled in the cache header - this is not needed and also breaks hermetic build contexts (a compiled file should _always_ result in the same output if Surelog didn't change - independent of time).
   *  Use range-based loop on flatbuffer content to improve readability.

### DIFF
--- a/include/Surelog/Cache/Cache.h
+++ b/include/Surelog/Cache/Cache.h
@@ -76,14 +76,18 @@ class Cache {
       flatbuffers::FlatBufferBuilder& builder, std::string_view schemaVersion,
       const std::filesystem::path& origFileName);
 
-  // Store errors and symbol tables in cache. Using a symbol table
-  // specifically for the cache, which is build and updating "cacheSymbols"
-  std::pair<flatbuffers::Offset<VectorOffsetError>,
-            flatbuffers::Offset<VectorOffsetString>>
+  // Store errors in cache. Canonicalize strings and store in "cacheSymbols".
+  flatbuffers::Offset<VectorOffsetError>
   cacheErrors(flatbuffers::FlatBufferBuilder& builder,
               SymbolTable* cacheSymbols,
               const ErrorContainer* errorContainer,
               const SymbolTable& localSymbols, SymbolId subjectId);
+
+  // Given the symbol table (that we've accumulated while emitting cached
+  // items), convert it to a flatbuffer cache vector.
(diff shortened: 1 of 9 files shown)


hzeller:
This will not fix yet the definition in a package missing as discussed offline, but it provides a cleaner base (less unrelevant data in the cache) to see what is happening.

alaindargelas:
I added a new test to prove the -nohash system works.
I think you might be missing the -nocomp flag when you are parsing independently the files and are getting an unknown type error. If you add -nocomp for the parsing stage, the error goes away and a link time (last command) all types are found:
  
https://github.com/chipsalliance/Surelog/pull/2906/files#diff-f075332fa3043f3d4be4e23fd2792a1233238ccee28139b0fd768d858b008113

alaindargelas:
@hzeller, I'm not sure if there is something wrong with my machine, but I think we just lost 30% runtime either with this change list #2902. Can you confirm?

hzeller:
mmh, so in my tests the regression consistently went from 1500 something to 1400 something CPU seconds, so it sped up things (I guess because less has to be written)

hzeller:
We already supply `-nocomp`; these are the options we give in the 'fill-the-cache' state
```
-noelab -nouhdm -nocomp -parse -nobuiltin -nohash -noinfo -nonote
```

In the fiinal 'link' stage, we leave out the `-noelab -nouhdm -nocomp`
```
--top-module foobar -parse -nobuiltin -nohash -noinfo -nonote
```

alaindargelas:
Using the testcase https://github.com/chipsalliance/Surelog/tree/master/tests/TestNoHash as a reference, can you try to reproduce the problem?
Is it a package enum that is used in a module or another package?

PR #2911, opened 2022-05-06
Resolve shell-utilities such as time and bash with /usr/bin/env

hzeller:
A strict posix system does not guarantee that the utilities
are available in /bin/bash or in the path, but should be
resolved by /usr/bin/env

While at it, make tcl regression more robust by choosing
the _last_ element in the SURELOG_COMMAND with lindex
instead of a hard-coded index (as this will change with
a double-element 'time' or even with a debug tool).


hzeller:
Merging as it does not touch anything that is likely in modification right now.

PR #2927, opened 2022-05-11
Build static libs relocatable

alaindargelas:
To make it easier to build the Yosys plugin


alaindargelas:
For the lamda Surelog/UHDM user, trying to make the build of the Yosys plugin easier.

alaindargelas:
Trying to get to:
cd Surelog
make
make install
cd yosys-f4pga-plugins
make LDFLAGS="-L/usr/local/lib/surelog -L/usr/local/lib/uhdm" << Even that should not be needed if the Makefile of the plugin was looking in /usr/local/lib too
make install

hzeller (dismissed):
LGTM!
Probably makes sense for all CI things to finish before merge.

hs-apotell:
Can you make this change in CMakeLists.txt instead of _forcing_ it via command-line thru' Makefile. For different platforms _relocatable_ has different meaning and the arguments to the compiler/linker are different which CMake takes care of. This configuration will apply to mac builds as well.

Issue #2928, opened 2022-05-11
Task reports a wrong begin


hs-apotell:
From builtin.cpp

```
  class mailbox;

    function new (int bound = 0);
    endfunction

    function int num();
    endfunction

    task put (message);
    endtask

    function try_put (message);
    endfunction

    task get (ref message);
    endtask

    function int try_get (ref message);
    endfunction

    task peek (ref message);
    endtask

    function int try_peek(ref message);
    endfunction

  endclass

  class process;

    typedef enum { FINISHED, RUNNING, WAITING, SUSPENDED, KILLED } state;

    static function process self();
    endfunction

    function state status();
    endfunction

    task kill();
    endtask

    task await();
    endtask

    task suspend();
    endtask

    task resume();
    endtask

  endclass
```

Reported UHDM:
```
...
  |vpiMethod:
  \_task: (work@mailbox::put), parent:work@mailbox
    |vpiName:put
    |vpiFullName:work@mailbox::put
    |vpiMethod:1
    |vpiVisibility:1
    |vpiIODecl:
    \_io_decl: (message), parent:work@mailbox::put
      |vpiDirection:1
      |vpiName:message
...
  |vpiMethod:
  \_task: (work@mailbox::get), parent:work@mailbox
    |vpiName:get
    |vpiFullName:work@mailbox::get
    |vpiMethod:1
    |vpiVisibility:1
    |vpiIODecl:
    \_io_decl: (message), parent:work@mailbox::get
      |vpiDirection:6
      |vpiName:message
...
  |vpiMethod:
  \_task: (work@process::kill), parent:work@process
    |vpiName:kill
    |vpiFullName:work@process::kill
    |vpiMethod:1
    |vpiVisibility:1
    |vpiStmt:
    \_begin: (work@process::kill), parent:work@process::kill
      |vpiFullName:work@process::kill
  |vpiMethod:
  \_task: (work@process::await), parent:work@process
    |vpiName:await
    |vpiFullName:work@process::await
    |vpiMethod:1
    |vpiVisibility:1
    |vpiStmt:
    \_begin: (work@process::await), parent:work@process::await
      |vpiFullName:work@process::await
  |vpiMethod:
  \_task: (work@process::suspend), parent:work@process
    |vpiName:suspend
    |vpiFullName:work@process::suspend
    |vpiMethod:1
    |vpiVisibility:1
    |vpiStmt:
    \_begin: (work@process::suspend), parent:work@process::suspend
[...]

PR #2935, opened 2022-05-14
Implement strong typedef for SymbolId & NodeId

hs-apotell:
Implement strong typedef for SymbolId & NodeId

* Define two global variables BadSymbolId & InvalidNodeId to mark invalid/initial state of SymbolId & NodeId respectively.
* In debug mode, SymbolId also holds a string_view representing the symbol itself. The only place exception where the id & symbol might get out of sync is during cache save/load because of the way the local & source SymbolTables are synchronized.
* For NodeId, 0 is no more a 'invalid' value. Since NodeId is used as an index into the array, using 0 as invalid value introduces edge case scenarios where calls to functions like Type, Line, etc become ambiguous.
* Indentified a few issues in the process (like #2928)
* Fixed a few const-ness issues
* Few fomatting/tidy changes
* Update clang-tidy config to force SymbolId & NodeId be used by value only, in constract to ref for objects, since for practical purposes these classes are mere wrappers around a native numeric type.
* Explicit typecast required to convert them back to raw types
* [BUGFIX] Typo error in CompileType.cpp:1702.


alaindargelas:
@hs-apotell, one conflicting log

hs-apotell:
@alaindargelas Resolved.

alaindargelas:
"For NodeId, 0 is no more a 'invalid' value. Since NodeId is used as an index into the array, using 0 as invalid value introduces edge case scenarios where calls to functions like Type, Line, etc become ambiguous."

0 was never an "invalid" value, for instance NodeId child = fC->Child(id), if child was 0, it meant there is no child (nullptr).
I don't like the whole initialization to Invalid part of this changelist. 0 meant nullptr, and init to 0 was way clearer to me. 

@hzeller to review.

alaindargelas:
The whole rename of NodeId to RawNodeId does not add to the readability.

hs-apotell:
> for instance NodeId child = fC->Child(id), if child was 0, it meant there is no child (nullptr)

And yet, calls to fC->Child(id) when id == 0 would end up walking the entire tree.

The intention here is to encapsulate what's valid within the NodeId class itself. The `operator bool` overload should be used to check for validity. Use of `InvalidNodeId` shouldn't be required at all. The only reason it exists is because the variables where initialized to zero previously and I just changed that to `InvalidNodeId`. I could have rather just took the initialization out all together and let the default constructor initialize it to the invalid value. I can clean this up in a follow up PR.

There's code like fC->Child(fC->Child(id)) without any checks on the first returned value from call to Child. In some cases, when NodeId == 0, the expectation was to start walking the tree to collect specific nodes of types (sl_collect functions). In other cases, when Node == 0, the expectation is to treat 0 as invalid and return 0 (Child, Type, Sibling, etc functions). This PR treats NodeId == 0, the same way across the board - a valid index. I found a number of places in code where the decisions are based on Type(0) calls which shouldn't be valid.

> The whole rename of NodeId to RawNodeId does not add to the readability.

Use of RawNodeId is restricted to Cache where the size of the native type becomes relevant and important. For Python API, NodeId can be exposed as a type to avoid using RawNodeId.

Feel free to propose alternative solutions.

hs-apotell:
Moreover, `InvalidNodeId` is not something newly introduced. It existed already and was defined as the magic constant 969696. The use of it however was limited in scope. If you would rather have 0 as invalid, one solution would be to initialize the FileContent::m_objects with an _empty/invalid_ value at index 0 and handle it accordingly across the board.

hzeller:
I like the idea of a strong type for these, I'll have a look tomorrow.

include/Surelog/Design/FileContent.h
```diff
@@ -68,15 +68,15 @@ class FileContent : public DesignComponent {
                    VObjectType type) const;  // Get first parent item of type
 
   NodeId sl_parent(
-      NodeId parent, const std::vector<VObjectType>& types,
+      NodeId parent, const std::unordered_set<VObjectType>& types,
```
hzeller, line 71: Don't mix two things in one PR: focus here is to change to the strong type not container choices.
So change this in a separate PR.

hzeller (commented):
Not through yet, but here a first round of comments.

include/Surelog/Common/SymbolId.h
```diff
@@ -3,10 +3,267 @@
...
+  bool operator>(const SymbolId &rhs) const { return id > rhs.id; }
+  bool operator>=(const SymbolId &rhs) const { return id >= rhs.id; }
+
+  bool operator==(const SymbolId &rhs) const { return id == rhs.id; }
+  bool operator!=(const SymbolId &rhs) const { return id != rhs.id; }
+
+ private:
+  RawSymbolId id;
+#if SYMBOLID_DEBUG_ENABLED
+  std::string value;
```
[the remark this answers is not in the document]
hs-apotell: This was a mistake. This was intended to be a string_view.

include/Surelog/Design/FileContent.h
hs-apotell: PR #2945

include/Surelog/Common/SymbolId.h
```diff
@@ -3,10 +3,267 @@
...
+#if SYMBOLID_DEBUG_ENABLED
+      value = rhs.value;
+#endif
+    }
+    return *this;
+  }
+
+  operator RawSymbolId() const { return id; }
+  operator bool() const { return id != BadRawSymbolId; }
+
```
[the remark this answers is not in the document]
hs-apotell: Includes only the ones that are actually used.

hs-apotell:
> 0 was never an "invalid" value, for instance NodeId child = fC->Child(id), if child was 0, it meant there is no child (nullptr).
I don't like the whole initialization to Invalid part of this changelist. 0 meant nullptr, and init to 0 was way clearer to me.

@alaindargelas  Revised. 0 is now the _nullptr_.
@hzeller Ready for review.

include/Surelog/Design/ModuleInstance.h
```diff
@@ -56,9 +56,9 @@ class ModuleInstance : public ValuedComponentI {
   void setInstanceBinding(ModuleInstance* boundToInstance) {
     m_boundInstance = boundToInstance;
   }
-  DesignComponent* getDefinition() { return m_definition; }
-  unsigned int getNbChildren() { return m_allSubInstances.size(); }
-  ModuleInstance* getChildren(unsigned int i) {
+  DesignComponent* getDefinition() const { return m_definition; }
```
hzeller, line 59: I recommend not using a const method when we return a non-const pointer, because it breaks the idea of `const`.

`const` indicates that the state of the object does not change, but since now this returns a non-const pointer, we can change the state of the object through the backdoor -- now even if we pass a const reference of `ModuleInstance`, we don't know if it is not modified because it is possible to retrieve a non-const pointer.

So only make a method const when it does not change the state of the object either directly or indirectly by returning a non-const pointer.

(if we have both use-cases, then I'd usually have two methods: 

```c++
const DesignComponent* getDefinition() const { ... }
DesignComponent* getMutableDefinition() { ... }
```

hzeller (changes requested):
You seem to change a lot of method to be a const signature that return some mutable state of the object. I've carefully made sure that this is not accidentally introduced so far, so we should not introduce it now :)

If there are places where we explicitly want to allow to get a mutable state, we should make this clear in signature and naming `const Foo* getFoo() const` and `Foo *getMutableFoo()` instead).

include/Surelog/Common/SymbolId.h
```diff
@@ -3,10 +3,266 @@
...
+  std::string_view value;
+#endif
+
+  friend std::ostream &operator<<(std::ostream &strm, const SymbolId &symbolId);
+};
+
+inline static const SymbolId BadSymbolId(BadRawSymbolId, BadRawSymbol);
+
+inline std::ostream &operator<<(std::ostream &strm, const SymbolId &symbolId) {
+  strm << (RawSymbolId)symbolId;
```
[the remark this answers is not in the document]
hs-apotell: Printing the string in the output would change the contents of the log between debug and release making it harder to diff.

include/Surelog/Design/ModuleInstance.h
hs-apotell: The 'const' here is for an accessory on 'ModuleInstance' whose state isn't changing because of this call. The state of 'Definition' however might change. Every _object_ dictates its own limits in polymorphism.

hs-apotell:
@hzeller Could we please finish the review on this PR. It's a large change (prone to merge conflicts) I don't want to sit on this for too long. I have made some arguments on the comments you made. Appreciate your time and attention.

include/Surelog/Common/SymbolId.h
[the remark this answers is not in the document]
hzeller: ok, if it is used in that context, we should leave it as-is.

include/Surelog/Design/ModuleInstance.h
hzeller: Yes, this is true to the letter, but not the spirit. This is why we should have const and non-const versions of the same thing, otherwise we'll never be able to fix the const-issues as returning non-const pointers from const getters will just plaster over the issue.

I suggest to not include these kind of const changes in this PR, as they  are unrelated.

include/Surelog/API/SLAPI.h
```diff
@@ -102,47 +102,47 @@ std::vector<antlr4::ParserRuleContext*> SLgetChildrenContext(
     SV3_1aPythonListener* prog, antlr4::ParserRuleContext* context);
 
 /* Parser API */
-NodeId SLgetRootNode(FileContent* fC);
+RawNodeId SLgetRootNode(FileContent* fC);
```
hzeller, line 105: (in a future change, these could probably all take a `const FileContent*` parameter. But not current scope)

include/Surelog/Design/ModuleInstance.h
hs-apotell: There are lot of const issues (and ownership problems) across the codebase.

Will revert the const changes where applicable.

hs-apotell:
@alaindargelas Github seems to be confused about the pending "Change Review" (happens sometimes if force pushed). The change is ready to be merged if CI passes.

P.S. Some configurations are failing in master as well and so are here.

hzeller (approved):

hzeller:
I've cleared the 'changes requested' flag.
@alaindargelas I think this looks good: the regression tests seem to fail similarly to the main branch currently. So once the logfiles are updated there, this probably will work as well.

PR #2945, opened 2022-05-18
Swap vector for unordered_set

hs-apotell:
Change FileContent API param from vector to unordered_set to avoid
having to iterate to find (the only purpose the input arg are used)

### DIFF
--- a/src/DesignCompile/ResolveSymbols.cpp
+++ b/src/DesignCompile/ResolveSymbols.cpp
@@ -65,13 +65,13 @@ void ResolveSymbols::createFastLookup() {
   // std::string fileName =  "FILE: " + m_fileData->getFileName() + " " +
   // m_fileData->getChunkFileName () + "\n"; std::cout << fileName;
 
-  std::vector<VObjectType> types = {
+  std::unordered_set<VObjectType> types = {
       VObjectType::slModule_declaration,    VObjectType::slPackage_declaration,
       VObjectType::slConfig_declaration,    VObjectType::slUdp_declaration,
       VObjectType::slInterface_declaration, VObjectType::slProgram_declaration,
       VObjectType::slClass_declaration};
 
-  std::vector<VObjectType> stopPoints = {
+  std::unordered_set<VObjectType> stopPoints = {
       VObjectType::slModule_declaration, VObjectType::slPackage_declaration,
       VObjectType::slProgram_declaration, VObjectType::slClass_declaration};
 
@@ -82,7 +82,7 @@ void ResolveSymbols::createFastLookup() {
     VObjectType type = m_fileData->Type(object);
     NodeId stId = m_fileData->sl_collect(object, VObjectType::slStringConst,
                                          VObjectType::slAttr_spec);
-    if (stId != InvalidNodeId) {
+    if (stId) {
       std::string name = SymName(stId);
       m_fileData->insertObjectLookup(name, object, m_errorContainer);
       std::string fullName = libName + "@" + name;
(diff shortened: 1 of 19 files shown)


third_party/tests/Google/Google.log
```diff
@@ -2971,12 +3049,22 @@ Processing: -cd chapter-16 -I../../../UVM/1800.2-2017-1.0/src/ ../../../UVM/1800
 
 [WRN:PP0113] ../../../UVM/1800.2-2017-1.0/src/macros/uvm_object_defines.svh:806:9: Unused macro argument "OP".
 
+=======
+>>>>>>> Update logs
```
hzeller, line 3053: this looks like a merge conflict header.

hzeller:
Why do the logs have to be updated ? If this is due to the hashing generated a different order, then you might want to use `std::set<>` instead of `unordered_set<>`

In many security-hardened environments, the hash function is randomized with each start of a binary so this will result in unpredictable tests in case this is due to the order.

hs-apotell:
> Why do the logs have to be updated ? If this is due to the hashing generated a different order, then you might want to use `std::set<>` instead of `unordered_set<>`
> 
> In many security-hardened environments, the hash function is randomized with each start of a binary so this will result in unpredictable tests in case this is due to the order.

Google test is reporting one additional error and the error itself seems very legitimate. Specifically, this particular error is reported with this error which isn't with current master.
From file `generic/typedef/typedef_test_28__bad.sv`.

third_party/tests/Google/Google.log
hs-apotell: Fixed.

hzeller:
I am wondering why the error is reported now and not before. Did you fix a bug with this change ?

src/DesignCompile/ResolveSymbols.cpp
```diff
@@ -65,13 +65,13 @@ void ResolveSymbols::createFastLookup() {
   // std::string fileName =  "FILE: " + m_fileData->getFileName() + " " +
   // m_fileData->getChunkFileName () + "\n"; std::cout << fileName;
 
-  std::vector<VObjectType> types = {
+  std::unordered_set<VObjectType> types = {
```
hzeller, line 68: These look like they can be 'static const', but that can probably  wait for another PR.

hzeller:
As a general remark, I'd probably typedef/using `using VObjectTypeSet = std::unordered_set<VObjectType>;` somewhere and use `VObjectSet`  everywhere where you now manually use `std::unordered_set<VObjectType>`.

Not only is it shorter to read, but it also makes it easier to play with different implementations.

hs-apotell:
> As a general remark, I'd probably typedef/using `using VObjectTypeSet = std::unordered_set<VObjectType>;` somewhere and use `VObjectTypeSet`  everywhere where you now manually use `std::unordered_set<VObjectType>`.
> 
> 
> 
> Not only is it shorter to read, but it also makes it easier to play with different implementations.

Agreed but unfortunately that hasn't been the convention in this code base. I can do a follow up pass if needed.

hs-apotell:
> I am wondering why the error is reported now and not before. Did you fix a bug with this change ?

Potentially. A few copy/paste errors in FileComtent.cpp would likely be the cause. Implicit conversion between VObjectType, NodeId and SymbolId was hiding these issues.

hzeller:
follow-up PR with the typedef/using sounds good to me.

hzeller:
LGTM.

Looks like indeed the FileContent.cpp actually does some more things correct now.
Leaving to @alaindargelas  to verify and acknowledge the new error in the log.

hs-apotell:
I will resolve any conflicts in the other PR once this is promoted to master. Feel free to comment on the other if anything else needs attention. I might try making zero as the InvalidNodeId since that was one of the concerns that @alaindargelas raised.

PR #2946, opened 2022-05-18
Add workflow for testing Yosys plugin support

rkapuscik:
This adds a separate CI workflow that:
* checks out current confirmed-to-work tool versions from https://github.com/antmicro/yosys-uhdm-plugin-integration
* builds the tools against current Surelog
* uses the tools to build Ibex as a reasonably complex test case

This aims to help catch regressions that break the SystemVerilog plugin for Yosys.

### DIFF
--- a/.github/workflows/yosys-plugin.yml
+++ b/.github/workflows/yosys-plugin.yml
@@ -58,6 +58,7 @@ jobs:
     - name: Build binaries
       run: |
         cd yosys-uhdm-plugin-integration
+        pip install cmake==3.22.4
         echo 'PATH=/usr/lib/ccache:'"$PATH" >> $GITHUB_ENV
         ./build_binaries.sh
 
--- a/.github/workflows/yosys-plugin.yml
+++ b/.github/workflows/yosys-plugin.yml
@@ -81,3 +81,10 @@ jobs:
       with:
         name: top_artya7.bit
         path: ./yosys-uhdm-plugin-integration/UHDM-integration-tests/build/lowrisc_ibex_top_artya7_surelog_0.1/synth-symbiflow/top_artya7.bit
+
+    - name: Upload load graphs
+      uses: actions/upload-artifact@v2
+      with:
+        name: plots
+        path: |
+          **/plot_*.svg
--- a/.github/workflows/yosys-plugin.yml
+++ b/.github/workflows/yosys-plugin.yml
@@ -58,8 +58,6 @@ jobs:
     - name: Build binaries
       run: |
         cd yosys-uhdm-plugin-integration
-        # ubuntu bionic have too old cmake, install one from pip
-        pip install cmake
         echo 'PATH=/usr/lib/ccache:'"$PATH" >> $GITHUB_ENV
         ./build_binaries.sh
 
--- a/.github/workflows/yosys-plugin.yml
+++ b/.github/workflows/yosys-plugin.yml
@@ -1,0 +1,79 @@
+name: 'yosys-systemverilog-plugin'
+
+on:
+  push:
+    branches:
+      - master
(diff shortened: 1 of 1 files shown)


kamilrakoczy (approved):
LGTM

mithro:
Please make sure there is something like this that is checking the latest versions, so we know when an change causes breakage.

mithro:
FYI - @hzeller

.github/workflows/yosys-plugin.yml
```diff
@@ -0,0 +1,79 @@
...
+
+    steps:
+
+    - name: Install dependencies
+      run: |
+        sudo apt-get update -qq
+        sudo apt install -y software-properties-common
+        sudo add-apt-repository ppa:ubuntu-toolchain-r/test
+        sudo apt-get update -qq
+        sudo apt install -y gcc-9 g++-9 build-essential cmake tclsh ant default-jre swig google-perftools libgoogle-perftools-dev python3 python3-dev python3-pip uuid uuid-dev tcl-dev flex libfl-dev git pkg-config libreadline-dev bison libffi-dev wget
```
hzeller, line 31: Surelog needs `python3-orderedmultidict` and `python3-psutil`, which are not mentioned here. Are they pulled in with `pip` ?

hzeller (approved):
... but double-check if we actually need to install an additional cmake.

.github/workflows/yosys-plugin.yml
```diff
@@ -0,0 +1,79 @@
...
+    - name: Use ccache
+      uses: hendrikmuhs/ccache-action@v1
+      with:
+        key: linux-yosys-submodules
+
+    - name: Build binaries
+      run: |
+        cd yosys-uhdm-plugin-integration
+        # ubuntu bionic have too old cmake, install one from pip
+        pip install cmake
```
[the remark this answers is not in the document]
rkapuscik: Looks like Focal Fossa [has cmake 3.16.3](https://packages.ubuntu.com/focal/cmake) and Surelog [needs >=3.13](https://github.com/chipsalliance/Surelog/blob/master/INSTALL.md#build-dependencies). Removed.

.github/workflows/yosys-plugin.yml
rkapuscik: `orderedmultidict` is installed through `pip` a few lines further. It seems that `psutil` is only needed for Surelog's regression tests and we're not running them here.

alaindargelas:
@rkapuscik, This is still a work in progress looks like? It's crashing.

hzeller:
@rkapuscik any update here ?

.github/workflows/yosys-plugin.yml
```diff
@@ -0,0 +1,90 @@
...
+        submodules: recursive
+        fetch-depth: 0
+        path: 'yosys-uhdm-plugin-integration/Surelog'
+
+    - name: Setup cache
+      uses: actions/cache@v2
+      with:
+        path: ${{ env.CCACHE_DIR }}
+        key: cache_${{ steps.cache_timestamp.outputs.time }}
+        restore-keys: cache_
```
kamilrakoczy, line 60: This requires ``cache_timestamp`` step, e.g.:
```
    - name: Create Cache Timestamp
      id: cache_timestamp
      uses: nanzm/get-time-action@v1.1
      with:
        format: 'YYYY-MM-DD-HH-mm-ss'
 ```

rkapuscik: Added, thanks!

kamilrakoczy (approved):

rkapuscik:
@hzeller looks like it consistently passes now, so should be ready for review.

PR #2971, opened 2022-05-27
Make snapshotting of SymbolTable explicit.

hzeller:
The copy constructor of the SymbolTable is creating
a snapshot of it; this change names it accordingly.

We do want to avoid accidental copying of a SymbolTable,
so the copy constructor is now private, requring to
call CreateSnapshot() when explicitly needed.

No functional change.


hzeller:
Context: The need for #2970 made it clear that it was prudent to make it hard to create accidental copies.

PR #2976, opened 2022-05-27
log update

alaindargelas:
scripts/regression.py   crashes. Had to update only this file


hzeller:
I added a fix to the regression script in #2973 so that should probably work now better.

PR #2980, opened 2022-05-27
IWYU: add some missing headers after NodeId/SymbolId header split.

hzeller:

### DIFF
--- a/include/Surelog/Design/Design.h
+++ b/include/Surelog/Design/Design.h
@@ -27,6 +27,7 @@
 
 #include <Surelog/Common/Containers.h>
 #include <Surelog/Common/NodeId.h>
+#include <Surelog/Common/SymbolId.h>
 
 #include <map>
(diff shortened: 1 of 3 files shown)


hs-apotell:
Didn't the CI build successfully? Just so I don't miss these next time - how did you find them missing??

hzeller:
Yes it compiled before in the CI, so it looked good there.

I tested if each header in itself is self-contained, so that discovered it. Maybe can add some CI using the [IWYU](https://include-what-you-use.org/) utility to determine these kind of things.

hs-apotell:
> add some CI using the [IWYU](https://include-what-you-use.org/) utility

Yes, we should. If you got time, can you provide the exact command to use for Posix environment. Even better if you could add a CI build for it that would be great.

hs-apotell:
Added #2981 to track.

Issue #2985, opened 2022-05-28
Print UhdmId by default in Uhdm output in logs


hs-apotell:
Any reason the default is set to false? The ID information help's track a lot of issues and has been very handy in identifying issues.

Can we toggle the flag as default?

alaindargelas:
Sure, it was very unstable at the beginning of the project since every time there is a new type of node or simply any new node added it shifts the IDs by the number of new nodes introduced. But things are stabilizing now. Add -d vpi_ids

hs-apotell:
Fixed by #2988

PR #2988, opened 2022-05-29
Issue #2985: Add object ids by default in uhdm output

hs-apotell:
Issue #2985: Add object ids by default in uhdm output

Also, fixed the regression script update logic to fix line-endings for Windows.

NOTE: PR includes [PR 723](https://github.com/chipsalliance/UHDM/pull/723) from UHDM

### DIFF
--- a/scripts/regression.py
+++ b/scripts/regression.py
@@ -304,6 +304,7 @@ def _run_surelog(
       rel_output_dirpath = rel_output_dirpath.replace('\\', '/')
 
     cmdline = ' '.join([part for part in parts if part] + [
+      '-d', 'vpi_ids',
       '-mt', str(mt),
       '-mp', '1' if '-lowmem' in cmdline else str(mp),
       '-o', rel_output_dirpath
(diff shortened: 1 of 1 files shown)


scripts/regression.py
```diff
@@ -745,24 +746,13 @@ def _update_one(params):
...
-      #
-      # with open(golden_log_filepath, 'wt', encoding='cp850') as ostrm:
-      #   for line in lines:
-      #     ostrm.write(line)
-      #     ostrm.write('\n')
-      #   ostrm.flush()
-      #   ostrm.close()
+      # On Windows, fixup the line endings
+      if platform.system() == 'Windows':
+        with open(golden_log_filepath, 'rt', encoding='cp850') as istrm:
```
hzeller, line 751: Wouldn't UTF-8 be more robust in case there is any unicode in the output ? cp850 sounds very 1990ies...

hs-apotell: That was immediate solution when I saw an encoding issue but that it didn't work. I remember the issue to being only on MSYS2.

That said, the logs have changed quite a bit and I remember you making a specific change to remove non-ascii characters from log.

I will try changing it to utf-8 across the script and see if anything falls. Will follow up with a new PR once the current open PRs are merged and dust settles.

alaindargelas:
This PR and https://github.com/chipsalliance/UHDM/pull/723 do not add-up in functionality.
I reverted https://github.com/chipsalliance/UHDM/pull/723 until there is more evidence that vpiParent info is visible in the visitor.

hs-apotell:
@alaindargelas Targeted change. Will reintroduce the vpiParent change as a separate PR.

alaindargelas:
Actually, this was a bad idea. I'm working on a change in UHDM elaboration, the IDs changes makes the diffs unreadable.

PR #2991, opened 2022-05-31
Handle -odir/--Mdir before processing arguments

rkapuscik:
This PR makes sure the output directory is set properly before processing arguments.

Without this the `-link` option always takes the default (`.`) folder (and fails if separate compilation was run elsewhere).


alaindargelas:
@rkapuscik, if this is ready, can you remove the Draft status?

PR #2993, opened 2022-06-01
Parenting in UHDM output

hs-apotell:
Parenting in UHDM output

Moving the parent information from being inlined to actually visiting
the parent. This is important for models like ref_obj which are unique
for each instance and carry important information like the 'actual'
parent and vpiActual.


alaindargelas:
@hs-apotell, you missed https://github.com/chipsalliance/Surelog/pull/2994
We cannot keep the IDs in the log, it makes diffing impossible.
Please turn off the id (Sync #2994 and #2995), rerun and resubmit your PRs.

alaindargelas:
@hs-apotell, this PR is not needed, I merged the UHDM changes.

PR #3023, opened 2022-06-08
Few improvements to VpiListener & removed dead code

hs-apotell:
Few improvements to VpiListener & removed dead code

* Introducing visited & callback to VpiListener (like UhdmListener)
  Also, merged the logic in vpi_listener into VpiListener itself to get
  access to both these containers easily in enter/leave callbacks.
* Removed parent and parentHandle parameters from VpiListener::enter/leave.
  The parameter itself was misleading since it was called 'parent' but
  it may not actually be the true parent of the object in the callback. The
  pointer was really the object right below the one on top of the callback
  hierarchy. The true parent is available via object->VpiParent() and access
  to the call hierarchy is available via callstack. Thus making both these
  parameters unnecessary.
* VpiListener also has a callback function (enter|leave)Any which gets
  called at most once for every visited object. For logic that's
  identical for a number of types, this is ideal place to implement it
  rather than having to override n number of functions and cloning the
  logic.
* Fixed up all VpiListener sub-classes like UhdmLint and SynthSubset
* Also fixed a few memory leaks in Surelog code.

### DIFF
--- a/src/DesignCompile/CompileHelper.cpp
+++ b/src/DesignCompile/CompileHelper.cpp
@@ -55,9 +55,9 @@
 #include <string.h>
 #include <uhdm/ElaboratorListener.h>
 #include <uhdm/ExprEval.h>
+#include <uhdm/VpiListener.h>
 #include <uhdm/clone_tree.h>
 #include <uhdm/uhdm.h>
-#include <uhdm/vpi_listener.h>
 
 #include <iostream>
 #include <string>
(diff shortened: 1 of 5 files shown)


hs-apotell:
@alaindargelas One test is failing `ClassFsm` but the difference seems justified. With this PR the `vpiActual` is resolving it correct but with the current master it is not.

PR #3024, opened 2022-06-08
Reduce memory leaks by purging serializer.

hzeller:
Previously, some leaks were addressed in the unit test by deleting UDHM objects directly. But since they are not the owners, this of course would result in a  double-delete. So also adapt the unit test.

This is the second time I added a manual call to `Serializer::Purge()` (the other was in the [UHDM Yosys Module](https://github.com/chipsalliance/yosys-f4pga-plugins/pull/343)).

Wouldn't it be good if the destructor of the [Serializer](https://github.com/chipsalliance/UHDM/blob/master/templates/Serializer.h#L75) had a call to Purge ? Or is there a reason why it doesn't ?


alaindargelas:
No reason, you can try.

PR #3027, opened 2022-06-09
Delete old blackbox test

hs-apotell:
Delete old blackbox test

There are two folders with the same name BlackBox and Blackbox. The old
one i.e. the latter seems to have been abandoned.


hs-apotell:
@alaindargelas Please confirm that this test isn't required and is abandoned. If needed and still valid, please discard this PR and create your own new one with a different name. Windows won't support two folders with the same name, different casing.

PR #3035, opened 2022-06-10
Don't use function that creates a large string first before writing to stream

hzeller:
Use stream function directly.

Context: https://github.com/chipsalliance/Surelog/issues/2962


hzeller:
Also context: https://github.com/chipsalliance/UHDM/pull/743#issuecomment-1152606857

hs-apotell (approved):
Can you update UHDM as well to remove my edit. It would be wise to just remove the function that returns a string.

hzeller:
Will do.

PR #3045, opened 2022-06-14
Code Maintenance/Cleanup Task

hs-apotell:
Code Maintenance/Cleanup Task

* Use StrCat/StrAppend where applicable
* Initialize variables at contruction time
* Moved constant strings as constexpr string_view
* Removed some duplicate code
* BUGFIX: Fixed a bug with parsing macro (-D) option
  (Line 761: def is always an empty string)
  (Line 792: def is once again always an empty string)
* Use local constexpr string_view instead of strings
* Use emplace instead of push_back
* Use std::stoi instead of atoi


src/CommandLine/CommandLineParser.cpp
```diff
@@ -1150,41 +1161,39 @@ bool CommandLineParser::parseCommandLine(int argc, const char** argv) {
...
       m_errors->addError(err);
       i++;
-    } else if (!all_arguments[i].empty() && all_arguments[i] == "--x-initial") {
+    } else if (all_arguments[i] == "--x-initial") {
       Location loc(mutableSymbolTable()->registerSymbol(all_arguments[i]));
       Error err(ErrorDefinition::CMD_PLUS_ARG_IGNORED, loc);
       m_errors->addError(err);
       i++;
-    } else if (!all_arguments[i].empty() && all_arguments[i].at(0) == '+') {
+    } else if (all_arguments[i].at(0) == '+') {
```
hzeller, line 1174: so here, the removed empty() check is probably still appropriate as we access a particular element in the string.
Or has it been ensured at this point that we can't have an empty arg ?

src/CommandLine/CommandLineParser.cpp
```diff
@@ -690,20 +701,24 @@ bool CommandLineParser::parseCommandLine(int argc, const char** argv) {
         m_debugCache = true;
       } else if (all_arguments[i] == "vpi_ids") {
         m_showVpiIDs = true;
-      } else {
-        int debugLevel = atoi(all_arguments[i].c_str());
+      } else if (all_arguments[i] == "coverelab") {
+        // Ignored!
+      } else if (is_number(all_arguments[i])) {
+        int debugLevel = std::stoi(all_arguments[i]);
```
hzeller, line 707: If we use `std::from_chars()` we can be compatible with a future in which all_arguments contains string-views.

hzeller (commented):
LGTM, but caveat in one place to make sure we never encounter an empty-string argument.

src/CommandLine/CommandLineParser.cpp
```diff
@@ -576,50 +593,59 @@ bool CommandLineParser::parseCommandLine(int argc, const char** argv) {
...
       }
     } else if (arg == "-l") {
       if (i < argc - 1) {
-        arg = StringUtils::unquoted(argv[i + 1]);
-        m_logFileId = m_symbolTable->registerSymbol(StringUtils::trim(arg));
-        m_logFileSpecified = true;
+        m_logFileId = m_symbolTable->registerSymbol(undecorateArg(argv[i + 1]));
       }
-    } else if (arg.find("-D") != std::string::npos) {
+    } else if (arg.find("-D") == 0) {
```
alaindargelas, line 627: -DFOO    
      and 
-DBAR=BLAH  

are both valid syntax, does your code work for both?

hs-apotell: Yes, this check is to ensure that the '-D' is at the start of the string and not somewhere in the middle. So, something like "-XABC-DEF" shouldn't succeed.

src/CommandLine/CommandLineParser.cpp
hs-apotell: Line 520 and then again on line 676.

src/CommandLine/CommandLineParser.cpp
hs-apotell: That future will likely need more changes than just swapping `std::stoi` to `std::from_chars()`. I switched to `std::stoi` because cl compiler then to warn about use of what it considers _unsafe_ functions.

PR #3046, opened 2022-06-15
Revert "Stop polluting the source folders during regression tests"

alaindargelas:
Reverts chipsalliance/Surelog#3042


alaindargelas:
@hs-apotell, this changelist completely breaks -mp option.
Please remind me what is this PR supposed to do? Why is it changing so many unrelated things at the same time?

hs-apotell:
> this changelist completely breaks -mp option.

All the regression tests have passed successfully across all platforms. This change doesn't do anything with `-mp` option. In fact, the number of changes in the `CommandLineParser.cpp` are minimal and repetitive. Which specific test reports problem with `-mp` option? Please provide more details.

> Please remind me what is this PR supposed to do?

Batch tests generated their outputs in the current folder. This change moves them to the output folder. The optional `cd` option in individual batch job is used as a relative path within the output folder. `-l` was incorrectly being parsed as a path rather than just a name of the file - which this change fixes.

> Why is it changing so many unrelated things at the same time?
What's unrelated here in this change. The CMake and Batch files that the `Compiler.cpp` generates _might_ have quoted strings in arguments and those weren't parsed correctly by `CommandLineParser.cpp` and so I fixed that.

alaindargelas:
cd Surelog/third_party/tests/black-parrot/bp_top/syn$ 
Edit runsurelog.sh and add -mp 8
./runsurelog.sh <path to Surelog>

With your change, there is only one process doing the parsing.
With current Surelog code, there are 8 processed managed by cmake that do the parsing, significantly speeding the parse phase.

I know none of that is working on Windows, so please remove all the changes you made in your PR that touches that code since you cannot test your changes.

hs-apotell:
So there is no test that verifies this option in the regression tests? I will verify the change in WSL.

hs-apotell:
Could you suggest a different test that should work with the `-mp` and doesn't require verilator setup? I can't seem to get verilator working under WSL.

alaindargelas:
You don't need Verilator to run this test.
Surelog acts as Verilator.
./runsurelog.sh  <path to Surelog>

PR #3065, opened 2022-06-20
RoundTrip Improvements

hs-apotell:
RoundTrip Improvements

* No more explicit elaboration
* Don't process builtin package
* Process only the non-elaborated tree (everything else discarded in enterAny)
* Typespecs are handled consistently rather than specializing certain use cases. Especially logic_typespec
* Handle missing location information gracefully

### DIFF
--- a/src/roundtrip.cpp
+++ b/src/roundtrip.cpp
@@ -48,7 +48,7 @@
 #undef interface
 #endif
 
-static constexpr char kOverwriteMarker = 0;
+static constexpr char kOverwriteMarker = '\0';
 
 typedef std::vector<std::string> file_content_t;
 typedef std::map<std::filesystem::path, file_content_t> design_content_t;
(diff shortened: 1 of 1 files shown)


src/roundtrip.cpp
```diff
@@ -3452,14 +3554,17 @@ static int run_in_uhdm_mode(int argc, const char **argv) {
...
-    }
-  }
+  // Do NOT elaborate explicitly since multiple designs are incomplete and
+  // an elaboration process for them doesn't end up well.
+  //
+  // for (const vpiHandle &handle : restoredDesigns) {
+  //   const UHDM::design *const design =
+  //       (const UHDM::design *)((const uhdm_handle *)handle)->object;
+  //   if (!design->VpiElaborated()) {
+  //     UHDM::ElaboratorListener listener(&serializer, false);
```
alaindargelas, line 3564: This should now be changed to UHDM::ElaboratorListener listener(&serializer, false, true);
if you need elaboration.

PR #3077, opened 2022-06-26
Include gate and udp instantiations in non-elaborated tree

hs-apotell:


hs-apotell:
@alaindargelas Need a little help in finishing this. Intent to have the gate and udp instantiations in the non-elaborated tree. I have managed to get most of it working. The issue I have is with setting `udp::Udp_defn`. Attached are the logs for the test that are impacted by this change.

[Build#846.zip](https://github.com/chipsalliance/Surelog/files/8987205/Build.846.zip)

* Should the value for `udp::Udp_defn` be valid in the non-elaborated model?
* For the elaborated model, should the value be set during elaboration using one of the bindXXX functions? I tried this but looks like the ElaboratorListener isn't tracking `udp` definitions. Also, should the definition itself be cloned for the elaborated model?

Appreciate some pointers to take this to finish line.

alaindargelas:
@hs-apotell, this is not coherent with the rest of the non-elaborated model. In the non-elaborated model, there are no instances of modules in modules. Why do you want instances of gates/udps? It is already present in the elaborated model (where it should be).
 


. Should the value for udp::Udp_defn be valid in the non-elaborated model?

  The non-elaborated model has many incorrect bindings (Because elaboration is typically required to do correct binding). But this one is one that can be done correctly without elaboration as udps don't have parameters.
The binding needs to happen in Surelog::UhdmWriter. Not in ElaboratorListener for the non-elaborated model. 

. For the elaborated model, should the value be set during elaboration using one of the bindXXX functions? I tried this but looks like the ElaboratorListener isn't tracking udp definitions. Also, should the definition itself be cloned for the elaborated model?

This is already done (in Surelog, not in ElaboratorListener):

Udp.log:
```
....
  |vpiPrimitive:
  \_udp: work@udp_body (work@udp_body_tb.udp), file:dut.sv, line:89:1, endln:89:22
    |vpiParent:
    \_module: work@udp_body_tb (work@udp_body_tb), file:dut.sv, line:84:1, endln:106:10
    |vpiDefName:work@udp_body
    |vpiName:udp
    |vpiFullName:work@udp_body_tb.udp
    |vpiUdpDefn:
    \_udp_defn: work@udp_body, line:1:1, endln:18:13
```

hs-apotell:
Elaboration is not an option for language server usage. User is working with just one file from a potentially larger design. Elaboration won't succeed for these use case. There are three options - 
* Use the non-elaborated model
* Fall back to the AST, or
* Use the antlr walker directly.

The non-elaborated model approach is little more useful in terms of functionality like go-to reference or declaration and definition. In my opinion, the non-elaborated model needs to be more complete with information on instances, comments, and macros (all of which is currently missing).

alaindargelas:
@hs-apotell, you can go ahead and implement in the non-elaborated model for all the items you mention.

hs-apotell:
> @hs-apotell, you can go ahead and implement in the non-elaborated model for all the items you mention.

That's exactly what I am attempting. I might come back to you with specific questions along the way.

Issue #3089, opened 2022-07-06
Noisy: [INF:CM0024] Executing with X processes and Y threads.


hzeller:
the `CMD_NUMBER_THREADS` is noisy and makes it hard to compare between machines as it depends on the number of cores.

Is this a message that should be as prominent ? Maybe it should only be there if `-profile` is on ?

PR #3090, opened 2022-07-06
Only emit process/thread count message if -profile is on.

hzeller:
Also, switch off -profile in a few tests that happen to have it
enabled.


hzeller:
Fixes #3089

hzeller:
How can I update the logfiles ?

The interesting thing is, some logfiles should change, but the `make test` actually doesn't pick it up, so it happens to work, but it shouldn't ?

alaindargelas:
python3 scripts/regression.py run
python3 scripts/regression.py update

Issue #2982, opened 2022-05-27
Add support for Clang on Windows


hs-apotell:
At the moment antlr doesn't build successfully with Clang on Windows. Need a few minor tweaks to the CMakeFile.txt to make it succeed.

Note: The goal is to support Clang/LLVM and NOT Clang-cl. The latter is M$ version of Clang that deviates heavily from standard.

hs-apotell:
@hzeller Could you assist in promoting [this](https://github.com/antlr/antlr4/pull/3678) PR to dev? These are few changes to antlr4 Cpp runtime CMakeFiles.txt to support building with Clang on Windows.

hzeller:
This looks like it is complete ?

hs-apotell:
Fixed by #3098 & #3100

PR #3098, opened 2022-07-10
Issue #2982: Add support for Clang on Windows

hs-apotell:
Issue #2982: Add support for Clang on Windows

Reduced tarball verbosity to make the logs sane!


hs-apotell:
@alaindargelas Test LoopParamOver is failing with infinite recursion. Seems to be happening only with Windows+Clang.

PR #3100, opened 2022-07-10
Fix WinClang stack overflow issue

hs-apotell:
Shorterm fix for WinClang stack overflow issue

The stack overflow was caused by the check loop functionality for
evaluating/reducing expressions. This needs a better logic than merely
depending on a stack frame counter.

Generally, for windows, the stack size has to be set for the application
but for some unknown/unidentified reason the stack usage for Clang seems
unnecessarily high. The set limit for the recursion was 1000 but even
with 64MB stack size the application kept running into the same
problem. Probably some other undelying issue (potential large objects
crated on stack) but a problem for some other day.
Implementing a short term solution to get the builds working again.

In the process, also found that the same problem happens with MSVC cl
compiler as well. In debug mode, the application blows up the stack.

Short term solution - For Windows only, reduce the limit from 1000 down to
100 and set the stack size to 32MB for clang builds.

A follow up problem - The value post reduction failure was invariably
dependent on the maximum loop counter. Even though the builds and tests
worked independently for the platform, the results were different across
platforms because of different max values for the loop counter. To
resolve this difference in resultant value post failure, use the input
unmodified.

### DIFF
--- a/src/DesignCompile/CompileExpression.cpp
+++ b/src/DesignCompile/CompileExpression.cpp
@@ -773,7 +773,8 @@ expr *CompileHelper::reduceExpr(any *result, bool &invalidValue,
   }
   expr *res =
       eval.reduceExpr(result, invalidValue, m_exprEvalPlaceHolder, pexpr);
-  return res;
+  // If loop was detected, drop the partially constructed new value!
+  return m_unwind ? nullptr : res;
 }
 
 any *CompileHelper::getValue(const std::string &name,
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -95,6 +95,9 @@ if(MSVC)
       "${CMAKE_CXX_FLAGS_RELEASE} ${TCMALLOC_COMPILE_OPTIONS} /Zc:__cplusplus /W0 /bigobj ${MY_CXX_WARNING_FLAGS}"
   )
   set(CMAKE_EXE_LINKER_FLAGS /STACK:8388608)  # 8MB stack size
+elseif(WIN32 AND (CMAKE_CXX_COMPILER_ID MATCHES "Clang"))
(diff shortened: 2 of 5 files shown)


alaindargelas:
Log file update needed

hs-apotell:
No that won't fix this issue. The log file is fine for all other configurations but The ones for windows. The difference in number of iterations for loop detection generates different outputs. Windows has 500 compared to other configurations where the maximum is 1000. The output needs to be predictable when a loop is detected. At the moment it is dependent on the number of iterations.

hs-apotell:
While at it, is it not possible to detect a loop by tracking the LHS, RHS and their corresponding locations? A second visit to the same name and location would mean a loop.

hs-apotell:
@alaindargelas Is the change in CompileExpression.cpp acceptable?? The reason to do this is to remove reduction's dependency on the max loop counter in the event when a loop is detected. Instead of using the last legal value, use the first i.e. from the original expression itself. This change will modify the logs but they will be consistent across the different configurations. Ideally, if a loop was detected the expression would be included unmodified but that seems harder to achieve with limited change.

src/DesignCompile/CompileExpression.cpp
```diff
@@ -773,7 +773,8 @@ expr *CompileHelper::reduceExpr(any *result, bool &invalidValue,
   }
   expr *res =
       eval.reduceExpr(result, invalidValue, m_exprEvalPlaceHolder, pexpr);
-  return res;
+  // If loop was detected, drop the partially constructed new value!
+  return m_unwind ? nullptr : res;
```
hs-apotell, line 777: @alaindargelas Is this acceptable change to break the dependency of the output on max loop counter?

PR #3135, opened 2022-08-04
shell.nix: use named channel instead of tar.

hzeller:
That way, the packages are always relative to the current distribution
and doesn't age away.

While at it, add ccache for faster compile time turnarounds.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -2,19 +2,12 @@
 # If you have nix installed, you may simply run `nix-shell`
 # in this repo, and have all dependencies ready in the new shell.
 
-{pkgs ? import (builtins.fetchTarball {
-  # Descriptive name to make the store path easier to identify
-  name = "nixos-2021-11";
-  # Commit hash
-  url = "https://github.com/NixOS/nixpkgs/archive/refs/tags/21.11.tar.gz";
-  # Hash obtained using `nix-prefetch-url --unpack <url>`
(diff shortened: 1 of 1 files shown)


hzeller:
@sjkelly to review (since you are the original owner of this shell.nix )
  * Is using the named package instead of a tar commonly used ? It should be more forward compabile and works on nixos and the nix package manager.
  * Adding ccache to the mix: while not really a dependency of Surelog, it certainly helps developer velocity. Is this the appropriate place to put it, or would we have another location than `buildInputs` be used ? (Noob here :) ).

shell.nix
```diff
@@ -2,19 +2,12 @@
...
-  # Descriptive name to make the store path easier to identify
-  name = "nixos-2021-11";
-  # Commit hash
-  url = "https://github.com/NixOS/nixpkgs/archive/refs/tags/21.11.tar.gz";
-  # Hash obtained using `nix-prefetch-url --unpack <url>`
-  sha256 = "162dywda2dvfj1248afxc45kcrg83appjd0nmdb541hl7rnncf02";
-}) {}}:
-
-
+{ pkgs ? import <nixpkgs> {} }:
```
sjkelly, line 5: This should be fine since I doubt any of these dependencies are changing in a meaningful way, as you see. Though it is not as robust to bitrot. Nix flakes should address the dependency pinning and composition problem a bit more robustly, and may be something to investigate in the near future.

sjkelly (approved):

PR #3147, opened 2022-08-11
Use .tar.gz instead of .tgz extension

hs-apotell:
Use .tar.gz instead of .tgz extension

Python's tar module messes things up with the .tgz extension making
unarchiving difficult due to the same file name used at multiple levels
of the folder hierarchy. Using .tar.gz (as recommended by the
documentation) seems to work as expected.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -110,14 +110,14 @@ jobs:
       run: |
         cd build
         mv regression surelog-linux-${{matrix.compiler}}-regression
-        find surelog-linux-${{matrix.compiler}}-regression -name "*.tgz" | tar czfp surelog-linux-${{matrix.compiler}}-regression.tgz -T -
+        find surelog-linux-${{matrix.compiler}}-regression -name "*.tar.gz" | tar czfp surelog-linux-${{matrix.compiler}}-regression.tar.gz -T -
 
     - name: Archive regression artifacts
       if: matrix.mode == 'regression' && always()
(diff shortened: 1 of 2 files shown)


.github/workflows/main.yml
```diff
@@ -110,7 +110,7 @@ jobs:
       run: |
         cd build
         mv regression surelog-linux-${{matrix.compiler}}-regression
-        find surelog-linux-${{matrix.compiler}}-regression -name "*.tgz" | tar czfp surelog-linux-${{matrix.compiler}}-regression.tgz -T -
+        find surelog-linux-${{matrix.compiler}}-regression -name "*.tar.gz" | tar czfp surelog-linux-${{matrix.compiler}}-regression.tgz -T -
```
hzeller, line 113: the file that is packed up then is named `*.tgz` again, wouldn't we want it name `*.tar.gz` as well ?

hs-apotell: Those are not archived using python. They are archived using GitHub action.

hzeller: Yes, but *.tar.gz is the correct naming anyway. *.tgz is a very uncommon way to name such archives.
Also, uniformly names are also beneficial as it leaves people less scratching their heads.

hs-apotell: Updated all references of .tgz to .tar.gz.

hzeller (approved):
LGTM

Let's wait for the CI to finish before merge.

PR #3149, opened 2022-08-12
Added flag for changing command line styles to support existing build flows from other tools

udif:
Here is a preliminary pull request, mostly for reviews (I haven't added any tests yet, only manual tests).
It supports a new '-style' flag with a mandatory parameter whose value must be either m, c or s.
At the moment it only adds '-incdir' , '-libext', '-define', '-liborder', '-librescan'.
I've manually tested '-define', and even that was not tested too deeply.

If the general direction here is OK, I'll clean this up and add automatic tests.

### DIFF
--- a/include/Surelog/CommandLine/CommandLineParser.h
+++ b/include/Surelog/CommandLine/CommandLineParser.h
@@ -229,6 +229,7 @@ class CommandLineParser final {
   CommandLineParser(const CommandLineParser& orig) = delete;
 
   bool plus_arguments_(const std::string& s);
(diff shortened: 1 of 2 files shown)


alaindargelas:
@udif, for backward compatibility reasons, please make sure that if -style is not specified, the current S style is used.

udif:
That's the case already.
In fact, I've defined "Default", as well as S, C and M, so I'm not even assuming the default is S.
At the moment I'm only adding aliases to the C style, and ignoring the rest.

Issue #2861, opened 2022-04-15
IncludeFileInfo is duplicated with cache enabled


hs-apotell:
Test: PreprocFunc

Logs from two consecutive runs attached.
[run1.log](https://github.com/chipsalliance/Surelog/files/8498911/run1.log)
[run2.log](https://github.com/chipsalliance/Surelog/files/8498910/run2.log)

Notice the IncludeFileInfo dump at the top of the files. The number of entries in the second log is twice as many as in first.

PR #3150, opened 2022-08-12
Issue #2861: IncludeFileInfo was being duplicated when cache is enabled

hs-apotell:
Issue #2861: IncludeFileInfo was being duplicated when cache is enabled

Clear the IncludeFileInfo container if using PP cache to allow the walk
to populate it again. Also, clear the IncludeFileInfo before popilating
it from cache (there's a default element in the container which gets
added from the PreprocessFile constructor).

### DIFF
--- a/include/Surelog/Cache/PPCache.h
+++ b/include/Surelog/Cache/PPCache.h
@@ -46,10 +46,10 @@ class PPCache : Cache {
   std::filesystem::path getCacheFileName_(
       const std::filesystem::path& fileName = "");
   bool restore_(const std::filesystem::path& cacheFileName, bool errorsOnly,
-                bool skipLineTranslationInfo);
+                int recursionDepth);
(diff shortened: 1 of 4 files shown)


alaindargelas:
Would have to debug the PR with the test failure.
What is the failure?
There is no log update with your PR.

alaindargelas:
@hs-apotell, simply use the same logic in PPCache:
```
  /* Restore include file info */
  CommandLineParser* clp = m_pp->getCompileSourceFile()->getCommandLineParser();
  if ((!clp->parseOnly()) && (!clp->lowMem())) {
    m_pp->clearIncludeFileInfo();
  }
```
and all the tests pass.

In -lowmem which you cannot emmulate on Windows, several pass are created with several execution of Surelog (some with -parseonly or -lowmem).

PR #3173, opened 2022-08-22
Fill typespec info for ports

hs-apotell:
Fill typespec info for ports

P.S. Need logs update.


alaindargelas:
@hs-apotell, please submit the log updates along with your PR. 
There is no way I can see if there is any problem without the log updates.

PR #3177, opened 2022-08-25
Tell clang-tidy that the complained-about line is ok.

hzeller:


alaindargelas:
Code formatting does not like that :-)

hzeller:
yeah, just fixing that

PR #3178, opened 2022-08-25
Provide a fallback path for a different version of clang-tidy.

hzeller:
That way, it is easier to use whatever version is available while
still using the preferred version if found.


hzeller:
Observed failure is just complaining about the thing fixed in #3177

PR #3179, opened 2022-08-25
Prefer empty() to size() when quering containers.

hzeller:
Depending on the container, size() might not be O(1). Also empty()
improves readability.


hzeller:
Failure is just clang-tidy complaining about the existing thing (that is fixed already in master)

PR #3182, opened 2022-08-26
Give user a choice to use flatbuffers from local system.

hzeller:
Providing SURELOG_USE_HOST_FLATBUFFERS to cmake.
Default OFF. If ON, flatbuffers from third_party/ is
skipped but rather attempted to be searched on the
host system using pkg-config.

This is the first step, similar options need to be
provided for capnp, antlr and uhdm.

This will make it much easier for packagers (e.g. Debian)
to lego-assemble the relevant components fitting into the
system.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,6 +18,16 @@ option(
   "If buildling with clang++ and libc++(in Linux). To enable with: -DWITH_LIBCXX=On"
   OFF)
 
+option(
+  SURELOG_USE_HOST_FLATBUFFERS
+  "Use flatbuffers library from host instead of third_party" OFF)
+
+# TODO SURELOG_USE_HOST_CAPNP
+# TODO SURELOG_USE_HOST_ANTLR
+# TODO SURELOG_USE_HOST_UHDM
+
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -607,13 +628,39 @@ endif()
...
   TARGETS surelog-bin
   RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
+
+# Libraries
+install(
+  TARGETS surelog
+  EXPORT Surelog
+  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}/surelog
+  PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/Surelog)
+
```
hzeller, line 638: For some of the following libraries, the question is if we even need to install them as they are static libraries. I guess it might depend on system-specific link rules if they are needed 'in production'. Something to worry about in follow-up cleanups.

hzeller:
Assigning to @hs-apotell for review, as you have a good understanding of cmake and can see if what I am doing here looks good.

PR #3199, opened 2022-09-02
gcc version

alaindargelas:


hzeller:
It looks like we actually do use [gcc 9](https://github.com/chipsalliance/Surelog/blob/51693c8f451c1334a955f49e0b0412551eddd9c8/.github/workflows/main.yml#L41) in the CI, so it should actually work with that older compiler as well.

What was the exact situation in which something didn't work well ? Could it have been a mix in compilers ?

alaindargelas:
https://github.com/antmicro/yosys-systemverilog/issues/1041#issuecomment-1234965084

PR #3200, opened 2022-09-02
Implement interface_array & module_array

hs-apotell:
Implement interface_array & module_array

Dependent on UHDM PR chipsalliance/UHDM#790


alaindargelas:
Interface arrays and module arrays do not help anything in the elaborated model, they are like keeping the model folded.
At most, they have their place in the non elaborated model. They have no place in the elaborated model.
We need the bit blasted view like we bit blast the generate statements loops.

tests/UnitElab/UnitElab.log
```diff
@@ -17362,6 +17427,117 @@ design: (work@bottom1)
         \_logic_net: (work@top.inst1.c), line:20:25, endln:20:26
       |vpiTypedef:
       \_logic_typespec: , line:22:8, endln:22:8
+  |vpiModuleArray:
```
alaindargelas, line 17430: This is redundant information

alaindargelas: For Generate statements we do not keep any folded view, we bit blast, same with the gate arrays, we have to be consistent. You can keep the module_arrays and instance_arrays in the non-elaborated model.

alaindargelas: Plus tools like Yosys and Verilator would see 2 views of the same thing, errors in perspective.
This PR cannot go in as is

alaindargelas:
I can't reconcile the explanation here:
https://github.com/chipsalliance/Surelog/issues/3198#issuecomment-1236051767

and the actual logs in the PR, the 2 don't match. I see a bunch of new entries that are redundant in the logs you added to the PR.

alaindargelas:
> > Do they have the same ID as the actual module instance parented to the parent module?
> 
> Yes, they have the same ID. The entries you have shown in your comment are elements of module_array/Modules container. This container holds references to the instantiated modules after blasting the array.
> 
> This is happening in UHDMWriter.cpp:L3493-L3510 here in this change.


OK, got it now. Thanks for the detailed explanation. I'll merge.

alaindargelas:
@hs-apotell update UHDM in this branch, I commented the non-standard extension -> conflict in UHDM

PR #3210, opened 2022-09-06
Add new APIs to SymbolTable

hs-apotell:
Add new APIs to SymbolTable

+ get(symbol) - returns a pair of symbol-id & pinned symbol
+ add(symbol) - returns a pair of symbol-id & pinned symbol
+ copyFrom(id, SymbolTable) - copies input to this table

### DIFF
--- a/include/Surelog/SourceCompile/SymbolTable.h
+++ b/include/Surelog/SourceCompile/SymbolTable.h
@@ -35,11 +35,10 @@
 #include <vector>
 
 namespace SURELOG {
-
 class SymbolTable final {
  public:
   SymbolTable();
-  ~SymbolTable();
+  ~SymbolTable() = default;
 
   // Create a snapshot of this symbol table. The returned SymbolTable contains
   // all the symbols this table has and allows to then continue using the new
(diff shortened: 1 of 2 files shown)


hzeller:
It looks useful, but where would it be used ? Can you describe/link to potential future uses ?

hzeller:
In particular then it would be good to know the trajectory.

include/Surelog/SourceCompile/SymbolTable.h
```diff
@@ -50,16 +49,21 @@ class SymbolTable final {
   // Register given "symbol" string as a symbol and return its id.
   // If this is an existing symbol, its ID is returned, otherwise a new one
   // is created.
-  SymbolId registerSymbol(std::string_view symbol);
+  std::pair<SymbolId, const std::string&> add(std::string_view symbol);
```
hzeller, line 52: In an effort to reduce std::string and thus have a way in the future to use different memory allocation: wouldn't returning a `std::pair<SymbolId, std::string_view>` be more desirable ?

Our (yours and my) goal is to eventually have `getSymbol(SymbolId)` return a `std::string_view` as well, so introducing now a new API that returns a `const std::string&` makes the work towards that future harder.

include/Surelog/SourceCompile/SymbolTable.h
```diff
@@ -50,16 +49,21 @@ class SymbolTable final {
   // Register given "symbol" string as a symbol and return its id.
```
hzeller, line 49: This comment needs updating: Returns a the SymbolId and the canonicalized string.

hzeller (commented):
That sounds good. It does not explain yet why we need the return of the [SymbolId + canonicalized string] though.

Anyway, I think I am happy to see how this unfolds, but we should probably return a `std::string_view` instead of a `const std::string&`

hzeller (approved):

PR #3219, opened 2022-09-09
IO Abstraction: Introducing PathId & FileSystem (Phase 1)

hs-apotell:
IO Abstraction: Introducing PathId & FileSystem (Phase 1)

NOTE: This is only the first phase of the large goal of IO abstraction.

Big picture goal is to abstract IO operations out of Surelog and into a class implementation that can be overriden by client application. This facilitates implementation of other more extensive features (like support for UNC paths, portale cache, multi-source root directories, cache archives, etc). Adding those directly to Surelog is not ideal since in many cases this features pull in other thrid-party dependecnies that might not have the same licensing terms as Surelog (and so leave it to the client/application to decide).

Introducing two new classes - PathId & FileSystem

PathId - New class parallel to SymbolId that represent path-like objects for IO operations. Each instance can represent a directory or a file but the internal representation doesn't necessarily have to be a _path_. For instance, in case of compressed archive of sources, it can be a string representation of the archive + the offset. The only enforcement is that the representation needs to be *printable* i.e. should be convertible to a string (since thee representation is stored in SymbolTable). For the advanced users. this limitation can very easily be overcome and is left as an exercise.

FileSystem - This reroutes access to native file system or in special advanced cases can restrict access to the file system. All operations of PathId are meant to be routed thru' this class. This commit contains only the bare-bones of the implementation (to keep the rest of the change review-able). Follow up commit will add the real meat to this implementation.

Most of this change is about switching from SymbolId to use of PathId where applicable. Some specific ones that may be of interest to the reviewer.

* Restrict use of std::filesystem to minimal scope (functions where possible). Do NOT pass std::filesystem::path across function boundaries. Use PathId instead.
* Do NOT cache unpinned strings, especially the ones that represent paths. Add them to the SymbolTable and resolve it on restore.
* PreprocessFile::m_fileId used to represent both a file path (when processing a file) or a macro name (when processing a macro). This was a problem since now the path needs to be represented using PathId but the macro name is represented using SymbolId. So, added two independent variables PreprocessFile::m_macroId & PreprocessFile::m_fileId.


hzeller:
There are absolutely no unit-tests that verify the new path-ID behavior, nor are there any extension to existing unit tests to verify corner cases.
Please for the next itereration, add unit also provide unit tests.

include/Surelog/Common/PathId.h
```diff
@@ -0,0 +1,140 @@
...
+#endif
+#endif
+
+namespace SURELOG {
+typedef uint32_t RawPathId;
+inline static constexpr RawPathId BadRawPathId = 0;
+inline static constexpr std::string_view BadRawPath = BadRawSymbol;
+
+class SymbolTable;
+
```
hzeller, line 30: This class is missing a description.

include/Surelog/Common/PathId.h
```diff
@@ -0,0 +1,140 @@
...
+      m_symbolTable = rhs.m_symbolTable;
+      m_id = rhs.m_id;
+#if PATHID_DEBUG_ENABLED
+      m_value = rhs.m_value;
+#endif
+    }
+    return *this;
+  }
+
+  const SymbolTable *getSymbolTable() const { return m_symbolTable; }
```
hzeller, line 65: It is unclear why a path-ID needs to know about the SymbolTable. It sounds like a leaky abstraction or 'convenience' member.
Ideally, this should be removed. Anything called 'ID' should not not carry more context around. Or it needs to be clearly explained in the PathId class description.

include/Surelog/Common/SymbolId.h
```diff
@@ -22,33 +23,28 @@ namespace SURELOG {
  *
  * Used to uniquely represent a string in SymbolTable. SymbolId can (and
  * should) be resolved only with the SymbolTable that it was generated with.
- *
+ *
```
hzeller, line 26: Trailing whitespace (if you configure your editor to use the `.editorconfig`, this can be automatic removed)

include/Surelog/Common/FileSystem.h
```diff
@@ -0,0 +1,70 @@
...
+#include <cstdint>
+#include <filesystem>
+#include <string_view>
+
+namespace SURELOG {
+class PathId;
+class SymbolId;
+class SymbolTable;
+
+class FileSystem {
```
hzeller, line 37: This needs a class description that describes the motivation and what it does.

include/Surelog/Common/FileSystem.h
```diff
@@ -0,0 +1,70 @@
...
+ public:
+  virtual PathId toPathId(const std::filesystem::path &path,
+                          SymbolTable *const symbolTable);
+  virtual const std::string &toSymbol(PathId id);
+  virtual std::filesystem::path toPath(PathId id);
+
+  virtual const std::filesystem::path &getCwd();
+  virtual PathId getCwd(SymbolTable *const symbolTable);
+
+  virtual PathId copy(PathId id, SymbolTable *const toSymbolTable);
```
hzeller, line 51: Given that this is a class called `FileSystem`, a function called `copy()` copies files around, right ?

If not, this needs a different name, and also, it needs a description.

Also, don't make the parameter `const`, it makes absolutely no difference (as you `const` the value, not the point location), and thus is just confusing.

include/Surelog/Common/FileSystem.h
```diff
@@ -0,0 +1,70 @@
...
+  static FileSystem *setInstance(FileSystem *const fileSystem);
+
+ public:
+  virtual PathId toPathId(const std::filesystem::path &path,
+                          SymbolTable *const symbolTable);
+  virtual const std::string &toSymbol(PathId id);
+  virtual std::filesystem::path toPath(PathId id);
+
+  virtual const std::filesystem::path &getCwd();
+  virtual PathId getCwd(SymbolTable *const symbolTable);
```
hzeller, line 49: Don't make the parameter `*const`, as it does not have any meaning but is just confusing
(it would have a meaning if the `const*` was left of the pointer).

Similar in other places of thie PR.

include/Surelog/Common/FileSystem.h
```diff
@@ -0,0 +1,70 @@
...
+class SymbolId;
+class SymbolTable;
+
+class FileSystem {
+ public:
+  static FileSystem *getInstance();
+  static FileSystem *setInstance(FileSystem *const fileSystem);
+
+ public:
+  virtual PathId toPathId(const std::filesystem::path &path,
```
hzeller, line 43: For something called `FileSystem` it does not contain anything that a file system would provide, like creating/removing/open/closing files.

Maybe the naming is off ? Are there plans to expand this ?

include/Surelog/Common/NodeId.h
```diff
@@ -117,4 +117,4 @@ typedef std::unordered_set<NodeId, NodeIdHasher, NodeIdEqualityComparer>
 
 }  // namespace SURELOG
 
-#endif  // SURELOG_NODEID_H
+#endif  // SURELOG_NODEID_H
```
hzeller, line 120: Final newline missing. Please configure your editor to use `.editorconfig`

include/Surelog/Common/PathId.h
```diff
@@ -0,0 +1,140 @@
+#ifndef SURELOG_PATHID_H
+#define SURELOG_PATHID_H
+#pragma once
```
hzeller, line 3: If you already have the include-guard, you don't need `#pragma once`. In fact, I'd probably leave it out as it is non-standard, so you can't rely on it working anyway. Include guards are the only one guaranteed to work.

include/Surelog/Common/PathId.h
```diff
@@ -0,0 +1,140 @@
+#ifndef SURELOG_PATHID_H
```
hzeller, line 1: there is a copyright header missing.

include/Surelog/Design/DesignComponent.h
```diff
@@ -78,7 +76,7 @@ class DesignComponent : public ValuedComponentI, public PortNetHolder {
   virtual unsigned int getSize() const = 0;
   virtual VObjectType getType() const = 0;
   virtual bool isInstance() const = 0;
-  virtual const std::string& getName() const = 0;
+  virtual std::string getName() const = 0;
```
hzeller, line 79: Wouldn't we want to strive for `std::string_view` instead ? You've now worsened the situation by changing a reference to a copy.

hzeller (commented):
A first round of review comments.

Unfortunately, this one is already merged, but before *any* other pull request can be added in this series, we first need to work out all the details here.

In general, this is too much and not reviewable, you really need to break independent changes to what they are

  * DesignComponent: change return value of name() from `const std::string&` to `std::string` can be a separate PR. In particular it would be then easy to discuss why this is not `std::string_view` and work towards what it is
  * PathId should start out as a no-op: have a typedef SymboldID to PathId, then change everything that needs to be a path from SymbolId to Pathid. Huge text change, but easy to review, no actual change.
  * All new classes need a description of what they are. Also naming: `FileSystem` sounds like a poor name for what it does, or it is not explained enough.
  * **All** new classes need unit tests.
  * Why does something that is called SometingId have a reference to the thing that supplies these IDs ? This sounds like a convenience hack.

Here are a first round of comments that we should address before the next phase of the filesystem series.

Whenever you work on changes, please make sure to break them up in simple steps
  * big changes only in a way that they are No-ops first, e.g. with the `typedef` of SymboldId -> NodeId.
  * introducing new classes one at a time with proper unit tests
  * Wire new classes up as needed in separate changes.

include/Surelog/Common/FileSystem.h
hzeller: Yes, but you typically would put that then in the _implementation_ part of the function, not the prototype. So in the API, you should keep it without the `*const`, in the implementation you then can choose to use it.

hzeller: `const*` tells the user that the input parameter is not changed. `*const` tells the function to not assign anything to the variable, but it doesn't make any difference to the caller.

Tthe 'const' definitely does not make a difference in the signature if it refers to the _variable_ being assigned.

The following signatures are identical:
```
  void foo(int x);
  void foo(const int x);
```

Sounds like your editor is overzealous, does simplified matching or wants to remind you of something that might be there in error.

include/Surelog/Common/PathId.h
hzeller: ah ok, yeah, let's remove it everywhere in a separate change.

include/Surelog/Common/FileSystem.h
hzeller: In which contexts is this used ? Maybe `duplicatePath()` could be a good name ?
What are the scenarios where this will be used ?

hzeller:
> To be honest, this isn't one night effort. This was a group effort from over three months. We did a number of design iterations for deciding how we wanted to implement some of the requested features in our tool. The change was organic but the effort paid off. I am cherry-picking from a much larger change in an effort to get this back into public repository.
> 
> Frankly, I thought that's what this change was - a big text change and doesn't change anything in terms of logic. No test case broke and no test log really change in any significant meaningful manner. All this change was doing is introducing PathId class in the codebase and fixing all the fallouts from it.
> 
> > PathId should start out as a no-op: have a typedef SymboldID to PathId, then change everything that needs to be a path from SymbolId to Pathid. Huge text change, but easy to review, no actual change.
> 
> That approach wouldn't have caught all the places that needed attention. I did do a few PR before this one for the changes that I could easily pull out from the larger effort. But breaking this into any smaller would have meant redoing (and verifying) a lot of work that was already completed.

But that is how code reviews and pull requests work.

Sure, you might not know beforehand what changes, that is why you work on a full change first until you're happy with it. The work then to make it reviewable is to break it up into the parts.
So now that you know what parts need to change to `PathId`, you then can retroactively do a change where everything is text-changed but made a typedef. etc.

Find all the changes independent from the rest, but that you also need for everything coming together, that can have their own review. Such as the `const std::string&` to `std::string` return value.

Yes, it requires a little bit more work to prepare a pull request, but that way it is not just a code-dump, but a well thought-out sequence of design decisions that the reviewers can follow.
I usually start out with the big local change (let's call it the creative phase), but then for review I re-do the exact same changes but break them out into separate commits.
That process often discovers that there are things that I don't have to change anymore.
Yes, it might take an hour of work, but it also introduces clarity and it is needed to convey ideas to fellow engineers. 

This code review contains 116 changed files! With a mix of functionality addition, text changes, return value changes...
That many files should only really be with no-op starting with a typedef (such as SymbolId -> FileId where needed), _or_ changing a single return value type throughout and have that an easy-to-review thing. So, large changes should only be text changes. Sometimes it requires a little 'dance': do a bulk change of something, then do some functionality change, do a bulk-change of some things turned back. And that is ok.

Any functionality change should be as minimal as possible (a few files and all the corresponding unit tests). Followed by a separate PR or at least commit that wires it up.

> 
> > DesignComponent: change return value of name() from const std::string& to std::string can be a separate PR. In particular it would be then easy to discuss why this is not std::string_view and work towards what it is
> 
> And I wouldn't have known that such a change would be needed unless I had the PathId change in place.

That is the point: you first develop things and find out what all is needed. That is for sure.
But _that_ is not the pull request. It is your responsibility to then break these up into independent, coherent changes before sending them as review. It also gives you another chance to reflect on why each change is needed (sometimes some change needed earlier in a larger effort turns out to not be needed anymore).

> All that said, I am happy to iterate on specific PR if you feel that needs more work and needs to be broken up. Unfortunately. given time constraints, it might not always be possible but will make best effort.

Please do. I am not saying that it is simple. And often it requires to re-do the change manually into multiple commits.
The burden  you put on Alain, me or others who have to review it is much higher; so you're transferring your time constraint to multiple times that effort to the reviewers. You see that Alain often just gives up on the review and merges, hoping for the best.

include/Surelog/Common/PathId.h
hzeller: leaking abstraction often makes implementation easier, but it doesn't mean it is the right thing.

I'll have a look at the CC file, but it currently smells like a convenience hack.

Issue #3221, opened 2022-09-09
Clean all warnings introduced


alaindargelas:
[ 79%] Building CXX object CMakeFiles/surelog.dir/src/SourceCompile/AntlrParserErrorListener.cpp.o
/home/alain/Surelog/src/DesignCompile/UhdmWriter.cpp: In member function ‘PLI_UINT32* SURELOG::UhdmWriter::write(SURELOG::PathId)’:
/home/alain/Surelog/src/DesignCompile/UhdmWriter.cpp:3611:36: warning: unused variable ‘fC’ [-Wunused-variable]
 3611 |           const FileContent* const fC = pf->getFileContent();
      |                                    ^~
[ 79%] Building CXX object CMakeFiles/surelog.dir/src/SourceCompile/AntlrParserHandler.cpp.o
[ 79%] Building CXX object CMakeFiles/surelog.dir/src/SourceCompile/CheckCompile.cpp.o
[ 80%] Building CXX object CMakeFiles/surelog.dir/src/SourceCompile/CommonListenerHelper.cpp.o
[ 80%] Building CXX object CMakeFiles/surelog.dir/src/SourceCompile/CompilationUnit.cpp.o
[ 80%] Building CXX object CMakeFiles/surelog.dir/src/SourceCompile/CompileSourceFile.cpp.o
[ 80%] Building CXX object CMakeFiles/surelog.dir/src/SourceCompile/Compiler.cpp.o
[ 80%] Building CXX object CMakeFiles/surelog.dir/src/SourceCompile/LoopCheck.cpp.o
[ 80%] Building CXX object CMakeFiles/surelog.dir/src/SourceCompile/MacroInfo.cpp.o
[ 82%] Building CXX object CMakeFiles/surelog.dir/src/SourceCompile/ParseFile.cpp.o
[ 82%] Building CXX object CMakeFiles/surelog.dir/src/SourceCompile/ParserHarness.cpp.o
[ 82%] Building CXX object CMakeFiles/surelog.dir/src/SourceCompile/PreprocessFile.cpp.o
[ 82%] Building CXX object CMakeFiles/surelog.dir/src/SourceCompile/PreprocessHarness.cpp.o
[ 82%] Building CXX object CMakeFiles/surelog.dir/src/SourceCompile/SV3_1aPpTreeListenerHelper.cpp.o
[...]

PR #3223, opened 2022-09-09
Issue #3221, #3219: Follow up and fixing warnings

hs-apotell:
Issue #3221, #3219: Follow up and fixing warnings

* Fixed build warnings and also updated the workflow to match local settings so the warnings should now be reported on CI as well.
* Add unit tests for PathId
* Some documentation & decorative changes


.github/workflows/main.yml
```diff
@@ -180,7 +180,7 @@ jobs:
         echo 'CXX=g++-9' >> $GITHUB_ENV
         echo 'PATH=/usr/lib/ccache:'"$PATH" >> $GITHUB_ENV
         echo 'PREFIX=${GITHUB_WORKSPACE}/install' >> $GITHUB_ENV
-        echo "ADDITIONAL_CMAKE_OPTIONS=-DPython3_ROOT_DIR=${pythonLocation} -DMY_CXX_WARNING_FLAGS='-W -Wall -Wextra -Wno-unused-parameter -Wno-unused-variable -Werror -UNDEBUG'" >> $GITHUB_ENV
+        echo "ADDITIONAL_CMAKE_OPTIONS=-DPython3_ROOT_DIR=${pythonLocation} -DMY_CXX_WARNING_FLAGS='-W -Wall -Wextra -Wno-unused-parameter -Werror -UNDEBUG'" >> $GITHUB_ENV
```
hzeller, line 183: Good! I wonder why that was there, maybe initially we did have a bunch of unused variables.

include/Surelog/Common/PathId.h
```diff
@@ -22,6 +38,19 @@
 #endif
 
 namespace SURELOG {
+/**
+ * class PathId
+ *
+ * Used to uniquely represent a file or directory abstractly.
```
hzeller, line 44: Does it represent a file/directory or a name of such ? It sounds like more a name which then can be used to open something that conceptually is a stream or directory.

include/Surelog/Common/PathId.h
```diff
@@ -22,6 +38,19 @@
...
+ * The context/value doesn't have to be a std::filesystem::path but
+ * can be any printable (i.e. convertible to string) value. This pinned
+ * value is stored in the PathId::m_symbolTable.
```
hzeller, line 47: It would probably be good to briefly describe why we don't store strings but rather IDs for size and uniqueness reasons.

hzeller: The fact that we keep the symbol table pointer is something I still disagree with, but I first have to look at the filesystem implementation attachment.
It indicates that the user of the `FileId` does not know which SymbolTable it belongs to.

Why do we even expect it to be in different SymbolTables ? It sounds like we want one SymbolTable in the filesystem implementation, where only the filesystem hands out new IDs. Since that is a singleton, it is also clear where that SymbolTable belongs to. That way, we also don't have the confusing mish-mash of actual symbols and filesystem-strings in the SymbolTable as it is now.

(we should probably rename SymbolTable at some point to string-unifier or SymbolFactory like in UHDM, because it is not really a SymbolTable, but that is for a separate change :) )

src/Common/PathId_test.cpp
```diff
@@ -0,0 +1,97 @@
...
+namespace fs = std::filesystem;
+
+namespace {
+class TestFileSystem : public FileSystem {
+ public:
+  TestFileSystem() : FileSystem(std::filesystem::current_path()) {}
+};
+
+TEST(PathId, constructors) {
+  std::unique_ptr<SymbolTable> symbolTableA(new SymbolTable);
```
hzeller, line 38: For readability, I suggest to have `new` calls on constructors look like function calls: `new SymbolTable()`

alaindargelas (approved):

PR #3230, opened 2022-09-15
IO Abstraction: Redirect read/write operations to FileSystem (Phase 2)

hs-apotell:
IO Abstraction: Redirect read/write operations to FileSystem (Phase 2)

Redirect all file read/write operations to go thru' the FileSystem. Also, includes API,

* to work with both text and binary formats.
* to access specific line within text files


alaindargelas:
@hzeller can you review this?

PR #3250, opened 2022-10-01
IO Abstraction: Redirect majority of FileUtils operations to FileSystem (Phase 3)

hs-apotell:
IO Abstraction: Redirect majority of FileUtils operations to FileSystem (Phase 3)

FileUtils::hashPath is an outlier. The function is soon to be deprecated in follow up PR and so not worth the effort of removing and fixing all the references.

Known Issue: Windows + Clang report CI failures. The configuration is disabled temporarily.


alaindargelas:
This PR is not acceptable in this shape.
It does touch too many things that are completely unrelated. Also why do we loose compiler support?

alaindargelas:
OK for removing the clang compiler on Windows. That is a compiler bug.

src/Utils/FileUtils.cpp
```diff
@@ -21,233 +21,24 @@
...
-      const fs::path& filepath = entry.path();
-      if (filepath.extension() == ext) {
-        result.emplace_back(fileSystem->toPathId(filepath, symbols));
-      }
-    }
-  }
-  return result;
-}
-
-std::vector<PathId> FileUtils::collectFiles(const fs::path& pathSpec,
```
alaindargelas, line 134: Where is this complex collect function implemented now?

PR #3252, opened 2022-10-02
Code maintenance

hs-apotell:
Code maintenance

* Use string reference wherever possible instead of cloning
* When parsing constants, look for prefixes not substring. i.e. use std::string::find == 0 instead of std::string::find == std::string::npos
* When looking for sign-ness allow both lower and upper case symbol consistently.

P.S. These are selected changes from other larger change and may not cover the specific case in its entirety.

### DIFF
--- a/src/Cache/ParseCache.cpp
+++ b/src/Cache/ParseCache.cpp
@@ -83,7 +83,7 @@ PathId ParseCache::getCacheFileName_(PathId svFileNameId) {
   }
   fs::path cacheDirName = fileSystem->toPath(cacheDirId);
   Library* lib = m_parse->getLibrary();
-  std::string libName = lib->getName();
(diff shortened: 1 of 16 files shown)


src/SourceCompile/AnalyzeFile.cpp
```diff
@@ -104,7 +104,11 @@ void AnalyzeFile::analyze() {
     std::string line;
     std::stringstream ss(m_text);
     while (std::getline(ss, line)) {
-      allLines.push_back(line);
```
alaindargelas, line 107: Please remove this change from the PR.
It has nothing do to with the rest of the PR. Please explain the rational here.

alaindargelas (changes requested):

PR #3264, opened 2022-10-09
Don't reduce when building non-elaborated tree

hs-apotell:
Don't reduce when building non-elaborated tree

When building non-elaborated tree some expressions in package where being reduced, which should happen only when building the elaborated tree.

### DIFF
--- a/src/DesignCompile/CompileHelper.cpp
+++ b/src/DesignCompile/CompileHelper.cpp
@@ -2531,6 +2531,9 @@ bool CompileHelper::isMultidimensional(UHDM::typespec* ts,
       bit_typespec* lts = (bit_typespec*)ts;
       if (lts->Ranges() && lts->Ranges()->size() > 1) isMultiDimension = true;
     } else if (ttps == uhdmstruct_typespec) {
+      // Intentionally treating struct as multi-dimensional to prevent
+      // reduction.
+      // https://github.com/chipsalliance/Surelog/pull/3264#discussion_r996568514
       isMultiDimension = true;
     }
   }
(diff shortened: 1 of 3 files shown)


alaindargelas:
The packages are compiled twice per your previous requirements.
First package is reduced:
https://github.com/chipsalliance/Surelog/blob/eacaa06b78c66f24440347137cd783a0d44a95e2/src/DesignCompile/CompilePackage.cpp#L46
Second is not:
https://github.com/chipsalliance/Surelog/blob/eacaa06b78c66f24440347137cd783a0d44a95e2/src/DesignCompile/CompilePackage.cpp#L51

It looks like your changes are impacting the 1rst compilation (topPackages) which it should not.
Your changes should only take effect on the allPackage

src/DesignCompile/CompileHelper.cpp
```diff
@@ -2530,8 +2530,6 @@ bool CompileHelper::isMultidimensional(UHDM::typespec* ts,
     } else if (ttps == uhdmbit_typespec) {
       bit_typespec* lts = (bit_typespec*)ts;
       if (lts->Ranges() && lts->Ranges()->size() > 1) isMultiDimension = true;
-    } else if (ttps == uhdmstruct_typespec) {
```
alaindargelas, line 2533: This is intentional to prevent reduction to happen. Please leave it the way it was.

alaindargelas (changes requested):

alaindargelas:
It's more complex than that.
It looks like that with your change a subsequent CompileHelper::expandPatternAssignment call is not working the way it is supposed to.

alaindargelas:
I'm looking into it.

Issue #3266, opened 2022-10-11
Multidimensional Parameters


hs-apotell:
Potential issue with multidimensional parameters

The value is computed here:
https://github.com/chipsalliance/Surelog/blob/24d778d0bc7741497597d2c217132c0bb07d7692/src/DesignCompile/CompileHelper.cpp#L2675

The value is forced to being true based on whether or not its compiling in elaboration mode (basically always true for non-elab mode):
https://github.com/chipsalliance/Surelog/blob/24d778d0bc7741497597d2c217132c0bb07d7692/src/DesignCompile/CompileHelper.cpp#L2706-L2708

The same value is stored in the parameter here (the value doesn't reflect the attributes of the parameter anymore since it was forced to be true for non-elab mode):
https://github.com/chipsalliance/Surelog/blob/24d778d0bc7741497597d2c217132c0bb07d7692/src/DesignCompile/CompileHelper.cpp#L2768

PR #3268, opened 2022-10-13
Issue #3266: Fix multidimensional parameters bug

hs-apotell:
Issue #3266: Fix multidimensional parameters bug

For non-elab tree, all parameters were being marked as multi-dimensional.


alaindargelas:
@hs-apotell, there is no log updates with this change? If yes, then please amend the PR with the log updates.

PR #3275, opened 2022-10-18
pattern assign const pushing

alaindargelas:


alaindargelas:
@hs-apotell this is my solution to your request

alaindargelas:
You get the idea now. I'll let you finish.
Please make sure you don't modify the elaborated tree. Or only to reduce it more.

PR #3292, opened 2022-10-28
Make preprocessor cache tolerant to unrelated changes

hs-apotell:
Make preprocessor cache tolerant to unrelated changes

Changes to include paths on command line shouldn't invalidate the cache if the additional includes (or otherwise) have no affect on the corresponding file.

Instead of caching include paths from command line, cache the include symbol from the source and the path that the symbol resolved to. For validation purpose, use the cached symbol to locate the included file again based on the current session's include paths and if the cached include path and the resolved include path match, the cache is still valid.

The include information was also unnecessarily being duplicated (as includes and include_file_info). Deprecated the former and using the include_file_info to check for cache validity.

Also, added a unit-test to test specifically for this scenario.


alaindargelas:
Please include log files update to see the impact on location

alaindargelas:
You changed code related to file and line location.
Please upload the logs

alaindargelas (changes requested):

PR #3294, opened 2022-10-28
Fix incremental builds with cache version change

hs-apotell:
Fix incremental builds with cache version change

Precompiled cache was not respecting version change since the check was completely being ignored. Ensure that the signature & version check is enabled for precompiled cache (both pp & parse cache).


alaindargelas:
Coredump in roundtrip:   | YosysSmallBoom              | PASS   |     0 |      0 |     0 |     291 |    296 |    21.37 |    3430 |    3085 |   0 **(-130714)/0 (-228458)** |

PR #3311, opened 2022-11-04
On unexpected filesystem errors: print out error in unittests.

hzeller:


hzeller:
The regression test issue is pre-existing, but the changed unittests are compiling and working as before.

PR #3321, opened 2022-11-08
FileContent::SymName now returns std::string_view instead of std::string

hs-apotell:
FileContent::SymName now returns std::string_view instead of std::string

Other file changes are ripple effect from the change to the FileContent::SymName API.

### DIFF
--- a/src/DesignCompile/CompileExpression.cpp
+++ b/src/DesignCompile/CompileExpression.cpp
@@ -82,8 +82,7 @@ bool CompileHelper::substituteAssignedValue(const UHDM::any *oper,
   return substitute;
 }
 
-any *CompileHelper::getObject(const std::string &name,
-                              DesignComponent *component,
+any *CompileHelper::getObject(std::string_view name, DesignComponent *component,
                               CompileDesign *compileDesign,
                               ValuedComponentI *instance, const any *pexpr) {
   any *result = nullptr;
@@ -285,7 +284,7 @@ any *CompileHelper::getObject(const std::string &name,
   return result;
 }
 
-UHDM::task_func *getFuncFromPackage(const std::string &name,
+UHDM::task_func *getFuncFromPackage(std::string_view name,
                                     DesignComponent *component,
                                     std::set<DesignComponent *> &visited) {
   for (Package *pack : component->getAccessPackages()) {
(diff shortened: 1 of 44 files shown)


src/DesignCompile/CompileExpression.cpp
```diff
@@ -433,13 +433,13 @@ static bool largeInt(std::string value) {
     largeInt = true;
   } else if (value.size() == 20) {
     if (isSigned) {
-      int64_t test = std::strtoll(value.c_str(), 0, 10);
+      int64_t test = std::strtoll(value.data(), 0, 10);
```
hzeller, line 436: this is invalid, as value is not nul terminated (also in other places in this PR).

Use `std::from_chars()` instead if you want to parse values from string-views.

hzeller (changes requested):
Using `data()` will not return a c-string.

src/DesignCompile/CompileExpression.cpp
hzeller: Yes, converting first to std::from_chars() and then doing the string_view change sounds good.

hzeller: ah, that can be because std::from_char is optional and older compilers (we're using gcc 9 in the CI) can't deal with it.
I think we need to have a fallback situation in that case.

I've implemented one in another project, maybe we need such a thing in Surelog
https://github.com/hzeller/beagleg/blob/main/src/common/string-util.cc#L82-L157

hzeller: ... and we do want to be able to use an slightly older compiler as many people in the EDA world unfortunately have pretty outdated machines.

hzeller:
Should I provide my implementation in `StringUtils.h` or maybe a new `NumberUtils.h` so that we can move forward with this change ?

(I think we're re-doing a lot of stuff with strings and numbers that we would get with [abseil](https://github.com/abseil/abseil-cpp), which we should probably consider at some point; another dependency, but very worthwhile)

hzeller:
Yes, strtoull() is probably a bit too lenient, parsing a negative value into an unsigned int. I have a pending #3336 providing some functions.

src/Expression/ExprBuilder.cpp
```diff
@@ -496,8 +496,8 @@ Value* ExprBuilder::evalExpr(const FileContent* fC, NodeId parent,
         break;
       }
       case VObjectType::slRealConst: {
-        const std::string& real = fC->SymName(child);
-        std::istringstream os(real);
+        const std::string_view real = fC->SymName(child);
+        std::istringstream os(real.data());
```
hzeller, line 500: This is not a valid use of `data()`
Since this is Why not use the new `parseDouble()` function here ?

hzeller: Sent #3392

PR #3331, opened 2022-11-10
Split regressions into shards.

hzeller:
First step: on linux.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   linux-build-test:
-    name: "${{ matrix.mode }}-${{ matrix.compiler }} Linux"
+    name: "test-${{ matrix.compiler }} Linux"
 
     runs-on: ubuntu-latest
     defaults:
@@ -15,16 +15,11 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        mode:
-        - test
-        - regression
-
         compiler:
         - gcc
         - clang
 
     env:
-      MODE: ${{ matrix.mode }}
       COMPILER: ${{ matrix.compiler }}
 
     steps:
@@ -57,7 +52,7 @@ jobs:
     - name: Use ccache
       uses: hendrikmuhs/ccache-action@v1.2
       with:
-        key: linux-${{ matrix.mode }}-${{ matrix.compiler }}
+        key: linux-test-${{ matrix.compiler }}
 
     - name: Configure compiler
       run: |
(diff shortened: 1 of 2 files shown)


hzeller:
Simplified version of the previous sharding approach. The main parts of the regression and valgrind runs are now around 45min.

Msys2 and MacOS have not been addressed yet.

More work needed
  * do the same thing for msys2 (which was defeating me so far), and MacOS.
  * Reduce boilerplate: since the regressions needed the sharding matrix, a bunch of boilerplate had to be copied over. This can probably be cleverly re-arranged to be more compact.

hzeller:
Yes, keeping it simple for now.

The rebuild is intentional to keep things simple in the first step (the previous experimental PR was changing too many things at once). There will be no walltime difference, as these run in parallel.

The failure shown in MacOS is addressed in #3332.

hzeller:
@proppy would you like to help simplify this file ?

.github/workflows/main.yml
```diff
@@ -88,37 +83,113 @@ jobs:
...
+      run: |
+        env
+        which cmake && cmake --version
+        which java && java -version
+        which python && python --version
+        which ninja && ninja --version
+        which $CC && $CC --version
+        which $CXX && $CXX --version
+
+    - name: Build
```
proppy, line 170: curious if you looked at reusing something like https://github.com/marketplace/actions/run-cmake ?

.github/workflows/main.yml
```diff
@@ -88,37 +83,113 @@ jobs:
...
+
+    - name: Run regression ${{matrix.shard}}/${{matrix.num_shards}}
+      run: |
+        python3 scripts/regression.py run --jobs $(nproc) --show-diffs \
+                                          --num_shards=${{matrix.num_shards}} \
+                                          --shard=${{matrix.shard}}
 
     - name: Prepare regression artifacts
-      if: matrix.mode == 'regression' && always()
+      if: always()
```
proppy, line 181: if you want to unify the shard artifact, you could use https://github.com/actions/cache to pass them a different dependent job with a unique key parameterized after `${{ github.run_id }}` (unique to the workflow invocation), and then restore and collect them together before uploading them as a single archive.

.github/workflows/main.yml
```diff
@@ -88,37 +83,113 @@ jobs:
...
+    - name: Cancel previous
+      uses: styfle/cancel-workflow-action@0.11.0
+      with:
+        access_token: ${{ github.token }}
+
+    - name: Install dependencies
       run: |
-        make test/regression
-        git status
+        sudo apt-get update -qq
```
proppy, line 120: use https://github.com/marketplace/actions/cache-apt-packages to speed thing up or build a local container image in another job marked as a dependencies of all the shards?

proppy (commented):
maybe @umarcor also have some ideas there?

hzeller:
Thanks @proppy, some good suggestions.I'll probably submit this version and then put in refactors in follow-up PRs.

Issue #3212, opened 2022-09-07
Food for thought: Replace UHDM::SymbolFactory with Surelog::SymbolTable implementation


hs-apotell:
There has been lots of iterations of improvements and optimizations on Surelog::SymbolTable but not much has changed for UHDM::SymbolFactory. Functionally both these classes are meant to be doing the same thing.

Is it worth moving the implementation of SymbolTable from Surelog and into UHDM and deprecating the UHDM::SymbolFactory. To make it less disruptive, UHDM::SymbolFactory itself could be a thin wrapper on top of SymbolTable so the API (and the client applications, if any) don't break.

Thoughts?

PR #3335, opened 2022-11-11
Issue #3212: Remove duplicate string table implementation

hs-apotell:
Issue #3212: Remove duplicate string table implementation

Move Surelog/SymbolTable to UHDM as SymbolFactory to replace its own implementation. Also, pulled in SymbolId implementation.

Updated code generation scripts to use SymbolId instead of SymbolFactory::ID.

IMPORTANT NOTE:
One important difference between UHDM's and Surelog's use of SymbolTable is that UHDM resolves an invalid SymbolId to an empty string but Surelog's expects the BAD_SYMBOL. So, the API in the implementation can't be the same. UHDM has its own set of APIs and Surelog it's own.


hzeller:
I am not sure about the benefits of unifying these at this time.

PR #3340, opened 2022-11-14
Provide a lenient parsing that allows full negative and unsigned int …

hzeller:
…parsing.

... the final 32 or 64 bits are then cast to the resulting integer.


hzeller:
It does.

PR #3342, opened 2022-11-15
CI: use custom runners for Linux tests

kgugala:
Signed-off-by: Karol Gugala <kgugala@antmicro.com>


kgugala:
@hs-apotell this can actually be detected so in forks we can run standard GH-runners. 

Basically this PR was to check how much runtime we can cut using faster machines.

alaindargelas:
@kgugala , please close if this PR has no merit anymore

PR #3380, opened 2022-12-13
unsized constant assign

alaindargelas:


src/DesignCompile/Elaboration_test.cpp
```diff
@@ -474,12 +475,14 @@ endmodule
       const std::string& name = cassign->Lhs()->VpiName();
       if (name == "p1") {
         // Val is 1, but it has a signed typespec (Meaning negative bin number)
-        EXPECT_EQ(val, 1);
+        EXPECT_EQ(val, 3);
```
hzeller, line 478: why did this value change ?

alaindargelas: Previous value was incorrect as unsigned where partially supported so far

alaindargelas: Now they get expended based on LHS size

src/DesignCompile/CompileHelper.cpp
```diff
@@ -2948,6 +2948,26 @@ bool CompileHelper::compileParameterDeclaration(
   return true;
 }
 
+char flip(char c) { return (c == '0') ? '1' : '0'; }
+
+std::string twosComplement(std::string bin) {
```
hzeller, line 2953: Always use std::string_view as parameter unless absolutely not possible.

PR #3383, opened 2022-12-15
Swap out std::strtoxxxx in favor of std::from_chars for portability

hs-apotell:
Swap out std::strtoxxxx in favor of std::from_chars for portability

Implemented NumUtils to contain all number related utility functions. Moved from global namespace to NumUtils class.

Known Issues:
Number parsing errors are largely ignored, to keep the behavior the same as previous implementation. Accounting for parse errors results in regression test failures. Will need to revisit these individually as a follow up.

### DIFF
--- a/include/Surelog/Utils/NumUtils.h
+++ b/include/Surelog/Utils/NumUtils.h
@@ -25,123 +25,167 @@
 #define SURELOG_NUMUTILS_H
 #pragma once
 
+#include <cctype>
+#include <charconv>
 #include <cstdint>
 #include <string>
-#include <charconv>
 #include <string_view>
-#include <ctype.h>
 
 namespace SURELOG {
-
-// These functions parse a number from a std::string_view into "result".
-// Returns the end of the number on success, nullptr otherwise.
-// So this can be used in a simple boolean context for success testing, but
-// as well in parsing context where advancing to the next position after the
-// number is needed.
-// All these functions require to check the return value to verify that parsing
-// worked.
-
-// Parse any number type with std::from_chars
-namespace internal {
-template <typename number_type>
-const char *strto_num(std::string_view s, number_type *result) {
-  while (!s.empty() && isspace(s.front())) s.remove_prefix(1);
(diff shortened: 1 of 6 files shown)


alaindargelas:
@hs-apotell there is a conflict to be resolved.

include/Surelog/Utils/NumUtils.h
```diff
@@ -25,123 +25,167 @@
...
+#include <cctype>
+#include <charconv>
 #include <cstdint>
 #include <string>
-#include <charconv>
 #include <string_view>
-#include <ctype.h>
 
 namespace SURELOG {
+class NumUtils final {
```
hzeller, line 35: I suggest to actually get _rid_ of the NumUtils class as using a static class just as a namespace is a bad pattern that we should get rid of.

So if you want these in a NumUtils namespace, create a namespace NumUtils (and also fix the previous NumUtils class).

src/DesignCompile/CompileExpression.cpp
```diff
@@ -488,7 +482,11 @@ constant *compileConst(const FileContent *fC, NodeId child, Serializer &s) {
         if (size.empty()) {
           c->VpiSize(-1);
         } else {
-          c->VpiSize(atoi(size.c_str()));
+          int32_t s = 0;
+          if (NumUtils::parseInt32(size, &s) == nullptr) {
```
hzeller, line 486: Instead of comparing with `nullptr`, the code would be more readable using it as bool. The fact that a pointer is returned only is useful for parsing functions that need to advance in a buffer. Using the pointer in expressions where you want to test if a number was parseable is decreasing readability.

So essentially almost everywhere you use `== nullptr` here, use the boolean equivalent. It makes the code _much_ more readable.

```
if (!NumUtils::parseInt(size, &s)) { ... }
```

include/Surelog/Utils/NumUtils.h
hzeller: Using a class as container of static functions is an antipattern that we should avoid and in particular not perpetuate further. Having an internal namespace is the lesser evil.

PR #3386, opened 2022-12-16
Issue chipsalliance/UHDM#831: Fix compile issues on MinGW

### LINKED ISSUE #3381: "module" is now reserved in C++
UHDM uses "module" as a model name. This doesn't build with C++20. Looking for suggestions on fixing this.

There are a few other names that also cause problems across platform. "interface" is reserved under Windows.

Following the convention used for other such scenarios, Would module_stmt & interface_stmt be acceptable?

Related issue: chipsalliance/UHDM#831

hs-apotell:
Issue chipsalliance/UHDM#831: Fix compile issues on MinGW

* Rename "interface" to "interface_inst" to avoid conflicting with system defined class with the same name.
* Rename "module" to "module_inst" since the latter is now a C++ reserved word as of C++ 20.

This also resolves #3381 

@alaindargelas This PR is dependent on and includes chipsalliance/UHDM#836

### DIFF
--- a/include/Surelog/Design/Netlist.h
+++ b/include/Surelog/Design/Netlist.h
@@ -47,7 +47,7 @@ class Netlist {
       InstanceMap;
   typedef std::map<std::string, UHDM::BaseClass*, std::less<>> SymbolTable;
 
-  std::vector<UHDM::interface*>* interfaces() { return m_interfaces; }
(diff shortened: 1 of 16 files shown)


alaindargelas:
@hs-apotell , I merged https://github.com/chipsalliance/UHDM/issues/831, please update this PR

PR #3401, opened 2022-12-23
Issue chipsalliance/UHDM#831: Fix compile issues on MinGW

hs-apotell:
Issue chipsalliance/UHDM#831: Fix compile issues on MinGW

* Rename "interface" to "interface_inst" to avoid conflicting with system defined class with the same name.
* Rename "module" to "module_inst" since the latter is now a C++ reserved word as of C++ 20.

PR Includes chipsalliance/UHDM#845


alaindargelas:
No there is no alternative. @rkapuscik to follow up with a fix

PR #3423, opened 2023-01-06
Replace main workflow with the sharded implementation

hs-apotell:
Replace main workflow with the sharded implementation

The sharded implementation has been working for a while. Good time to switch to reduce VM allocation since Github limits number of concurrently running VMs per account.


alaindargelas:
Conflict

PR #3505, opened 2023-03-13
Address an old TODO: replace NO_TCMALLOC=On with SURELOG_USE_TCMALLOC…

hzeller:
…=Off

The negative logic (setting a variable to ON to get the OFF) behavior in the tcmalloc macro was always a bit mind-twisting. So now formulate it as positive logic with the common SURELOG_USE_* prefix.
The default for tcmalloc is still on, so the default should behave the same.


hzeller:
Also @timkpaine as you know cmake well and might need to be aware of the option change since you're working on some packages.

hzeller:
Adding both @hs-apotell and @alaindargelas for review. I don't expect anything to break directly, but this flag might be baked into your personal workflow, so this is mostly a heads-up.

hzeller:
Sounds good to me. So `SURELOG_WITH_TCMALLOC` would be the name ?
 (I think it is good to qualify the project in front so that it is clear in a multi-project linkage which is meant)

hzeller:
We prefix the other variables with `SURELOG_`, (four `SURELOG_USE_HOST_*` variables) wouldn't it be inconsistent to put surelog in the middle in `WITH_SURELOG_*` ?

I am not following typical cmake recommendations, so if it is a strong preference in cmake to use `WITH_` as the prefix, then of course we should use `WITH_SURELOG_TCMALLOC`.

alaindargelas:
@hzeller you have a deterministic failure with this PR, I rerun the failed test and they fail again.
Please address (Or @hs-apotell since the failure only shows on Windows platform).

timkpaine:
Arrow mixes `WITH`, `USE`, and <empty>  https://github.com/apache/arrow/blob/main/cpp/CMakeLists.txt#L660

alaindargelas (changes requested):
@hzeller @hs-apotell, this PR fails the Windows release target.

.github/workflows/main.yml
```diff
@@ -711,13 +709,12 @@ jobs:
...
         set CC=clang
         set CXX=clang++
-        set NO_TCMALLOC=On
         set PREFIX=%GITHUB_WORKSPACE%\install
         set CPU_CORES=%NUMBER_OF_PROCESSORS%
 
         set MAKE_DIR=C:\make\bin
         set PATH=%pythonLocation%;%JAVA_HOME%\bin;%MAKE_DIR%;%PATH%
-        set ADDITIONAL_CMAKE_OPTIONS=-DPython3_ROOT_DIR=%pythonLocation%
+        set ADDITIONAL_CMAKE_OPTIONS="-DPython3_ROOT_DIR=%pythonLocation% -DSURELOG_WITH_TCMALLOC=Off"
```
hzeller, line 717: Is this the correct way to quote environment variables with spaces @hs-apotell ?

hzeller:
Changed to `SURELOG_WITH_TCMALLOC`.

I have no idea what the problems with the includes are. Maybe I have not quoted the environment variables correctly on Windows ? I used double quotes there, but maybe that needs to be done differently on the Windows shell. Maybe @hs-apotell can help ?

hzeller:
I am now trying without quotes on the Windows environment variable

.github/workflows/main.yml
hzeller: Ok, so removing the quotes helped.

PR #3510, opened 2023-03-15
Only compile std::regex once.

hzeller:
Make compiled static const so that they only have to be compiled the first time they are used, not every time.

While at it, improve readability of a regexp-string with backslash-quotes with raw-strings. Also fixes the clang-tidy warning.

### DIFF
--- a/src/SourceCompile/SV3_1aPpTreeShapeListener.cpp
+++ b/src/SourceCompile/SV3_1aPpTreeShapeListener.cpp
@@ -61,8 +61,8 @@ void SV3_1aPpTreeShapeListener::enterComments(
           ->getCommandLineParser()
           ->reportNonSynthesizable()) {
     if (ctx->One_line_comment()) {
-      std::regex reg1("\\/\\/\\s*synopsys\\s+translate_off\\s*");
-      std::regex reg2("\\/\\/\\s*pragma\\s+translate_off\\s*");
+      static const std::regex reg1(R"(\/\/\s*synopsys\s+translate_off\s*)");
+      static const std::regex reg2(R"(\/\/\s*pragma\s+translate_off\s*)");
       const std::string &text = ctx->One_line_comment()->getText();
       if (std::regex_match(text, reg1) || std::regex_match(text, reg2)) {
         m_filterProtectedRegions = true;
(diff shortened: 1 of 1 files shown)


src/SourceCompile/SV3_1aPpTreeShapeListener.cpp
```diff
@@ -61,8 +61,8 @@ void SV3_1aPpTreeShapeListener::enterComments(
           ->getCommandLineParser()
           ->reportNonSynthesizable()) {
     if (ctx->One_line_comment()) {
-      std::regex reg1("\\/\\/\\s*synopsys\\s+translate_off\\s*");
-      std::regex reg2("\\/\\/\\s*pragma\\s+translate_off\\s*");
+      static const std::regex reg1(R"(\/\/\s*synopsys\s+translate_off\s*)");
```
alaindargelas, line 64: I guess ChatGPT has to improve the readability of his code:

ChatGPT> write a C++ regexp to check for // synopysys translate_off
Here's an example C++ regular expression that can be used to match the string "// synopysys translate_off" in a C++ source code file:

c
Copy code
#include <regex>

std::string str = "// synopysys translate_off";
std::regex reg("\\/\\/\\s*synopysys\\s+translate_off\\s*");

if (std::regex_match(str, reg)) {
    // the string matches the regular expression
} else {
    // the string does not match the regular expression
}

Explanation:

#include <regex>: include the C++ regular expression library.
std::string str: declare a C++ string to hold the input string to be checked.
std::regex reg("\\/\\/\\s*synopysys\\s+translate_off\\s*"): declare a C++ regular expression that matches the string "// synopysys translate_off". The regular expression consists of:
\\/\\/: matches the two forward slashes that start the comment.
\\s*: matches any number of whitespace characters (spaces or tabs).
synopysys: matches the word "synopysys".
\\s+: matches one or more whitespace characters.
translate_off: matches the words "translate_off".
\\s*: matches any number of whitespace characters at the end of the line.
std::regex_match(str, reg): checks whether the string str matches the regular expression reg. Returns true if the string matches the regular expression, and false otherwise.

hzeller: Neat. Yeah, chatGPT should consider performance (`static const`...) and readability (raw strings), but pretty close.

PR #3519, opened 2023-03-22
Non-Synth change from ERROR to WARNING

alaindargelas:


alaindargelas:
@tgorochowik FYI, you can now use  -synth and not get errors, only warnings.

tgorochowik:
Thank you!

PR #3521, opened 2023-03-22
Derive the version number from the generated version header.

hzeller:
That way, we have a single source where the version number is coming from: the CMake project.


hzeller:
CC: @timkpaine using your generated header :)

PR #3523, opened 2023-03-22
pkg-config: provide right include base and add missing library.

hzeller:
In all examples, we include <Surelog/SomeHeader.h>, which means the pkg-config needs to point to the directory where the Surelog/ directory is installed.

Also: add the antlr4 runtime library needed (TODO: if we use SURELOG_USE_HOST_ANTLR, would this need to be replaced with a 'Requires antlr4' ?)

### DIFF
--- a/cmake/configs/Surelog.pc.in
+++ b/cmake/configs/Surelog.pc.in
@@ -1,10 +1,12 @@
 prefix="@CMAKE_INSTALL_PREFIX@"
 exec_prefix="${prefix}"
 libdir="${prefix}/lib/surelog"
-includedir="${prefix}/include/Surelog"
+includedir="${prefix}/include"
 
 Name: @PROJECT_NAME@
 Description: @CMAKE_PROJECT_DESCRIPTION@
 Version: @PROJECT_VERSION@
+# Currently, Surelog and UHDM version numbers go in tandem
+Requires: UHDM >= @PROJECT_VERSION@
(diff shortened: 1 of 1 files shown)


hzeller:
CC: @timkpaine as you have packaging experience: in the case the antlr4 libraries are not installed from Surelog, do we need to skip `-lantlr4-runtime` and instead talk about required libs ?

How about flatbuffers ?

hzeller:
mmh, so looks like antlr4-runtime does [not contain](https://packages.debian.org/sid/i386/libantlr4-runtime-dev/filelist)  a pkg-config file *.pc file, so adding it to our *.pc link library is probably best.

Flatbuffers: I am not entirely sure if we even need that at runtime.

timkpaine:
There's a lot of variability possible, but I think if we stick with the defaults (e.g. usually antlr installs in /usr/local, etc) it's good.

One other big issue I noticed, but did not fix, is capitalization and folders. UHDM puts its libraries in uhdm and includes in UHDM, Surelog puts its libraries in lower case surelog but includes in camel case Surelog. This is an easy but noisy change, since most source files need to change.

hzeller:
The includes base-dir in UHDM is already lowercase, so that is good.
The toplevel `Surelog/` as the toplvel surelog directory is a bit of a sore view, but I guess it goes together with the also CamelCase.h files...

PR #3530, opened 2023-03-23
Update UHDM to latest and bump Surelog to 1.49

hzeller:


hzeller:
FYI @timkpaine 
The auto-tagging will not work as we will have to figure out what it takes to tag from github actions, so I will just manually tag the version when pushed while figuring out what needs to be done to make it automatic.

PR #3538, opened 2023-03-24
install shared library in same location as static libraries

timkpaine:
for: https://github.com/chipsalliance/Surelog/pull/3536


hzeller:
Will `ldconfig` find it if it is in a subdirectory ? (probably yes, it looks like on a typical system there are lots of subdirectories with *.so files).

I think something similar needs to be applied to UHDM (also, there the location is hard-coded to `lib/uhdm`, while in surelog it is `${CMAKE_INSTALL_LIBDIR}/uhdm`

PR #3548, opened 2023-03-28
Canonicalize way to fill in pkg-config.

hzeller:


hzeller:
CC: @timkpaine This seems to be working better over a wider range of operating systems.

Issue #3541, opened 2023-03-24
strange install location of `uvm_pkg.sv.slpa`


hzeller:
On `make install`, the pre-compiled system verilog libs end up in `${PREFIX}/bin/pkg/work/`
```
ls -l bin/pkg/work 
total 23980
6169944 Mar 24 09:22 ovm_pkg.sv.slpa
1209576 Mar 24 09:22 ovm_pkg.sv.slpp
14381088 Mar 24 09:23 uvm_pkg.sv.slpa
2783720 Mar 24 09:22 uvm_pkg.sv.slpp
```

Which is a somewhat unusual location (or rather: "one should not do that" location :) ).

Typically, one would expect that in `${PREFIX}/share/surelog` (up to you if that should have a `pkg/` or `pkg/work/` subdirectory from there).

CC: @timkpaine who also has packaging experience.

timkpaine:
Agreed, I also noticed `pkg` ending up in `bin`, I would expect something like `share/surelog/` or `share/surelog/sv/` or similar

alaindargelas:
@hs-apotell, CMakeLists.txt and your new FileSystem class needs to be updated, can you take on that task?
I prefer the "share/surelog/pkg/" path as we are not saving .sv files but precompiled packages in binary format.

hzeller:
Currently, the install location is in  `${PREFIX}/bin/pkg/work`. So instead of the bin/ path, `${PREFIX}/share/surelog/pkg` is preferred.

alaindargelas:
Yes, ${PREFIX}/share/surelog/pkg/work  will work.

alaindargelas:
Fixed by #3556

PR #3556, opened 2023-03-30
Issue #3541: Change install path for precompiled data

hs-apotell:
Issue #3541: Change install path for precompiled data

* Switched to ${PREFIX}/share/Surelog
* Report cache usage in logs
* Update FileSystem to look for precompiled cache in the new location


timkpaine:
Need this one too https://github.com/chipsalliance/Surelog/blob/803b2ca762698e3bd2010d4fbc1302ad0120221e/cmake/configs/Surelog.pc.in#L3

hzeller:
The directories should not be upper-case, please leave them lowercase.

hzeller:
Looks mostly good.
It generates a double  'pkg/' , so the directory sturcture looks like that: `share/surelog/pkg/pkg/`, can this be just one ?

```
$ PREFIX=/tmp/share-test make install
$ find /tmp/share-test/ -name "*_pkg*"
/tmp/share-test/share/surelog/pkg/pkg/work/uvm_pkg.sv.slpp
/tmp/share-test/share/surelog/pkg/pkg/work/uvm_pkg.sv.slpa
/tmp/share-test/share/surelog/pkg/pkg/work/ovm_pkg.sv.slpa
/tmp/share-test/share/surelog/pkg/pkg/work/ovm_pkg.sv.slpp
```

hzeller:
Is there a way we can invoke surelog that lists all the pre-installed packages it found ? We can use that as a litmus test in the CI to make sure after an install these things are found.

alaindargelas:
Fixed #3556

hzeller:
Was the problem of the doubled-up `pkg/` addressed before the merge ? (`share/surelog/pkg/pkg/`)

PR #3558, opened 2023-03-31
Issue #3541: Fix path computation for installed precompiled data

hs-apotell:
Issue #3541: Fix path computation for installed precompiled data


hzeller:
Looks better, it does not cause a repeated pkg/, but now the unnecessary 'work' subdirectory is in there

```
${PREFIX}/share/surelog/pkg/work/ovm_pkg.sv.slpa
```

`work/` sounds like a build artifact, but not something that would add meaningful information to a installation location.  

So maybe better

```
${PREFIX}/share/surelog/pkg/ovm_pkg.sv.slpa
```
?

PR #3577, opened 2023-04-10
Add a project scope config header

hs-apotell:
Add a project scope config header


alaindargelas:
@hs-apotell , please fix the conflict.

PR #3585, opened 2023-04-12
Default value for port

alaindargelas:


src/DesignCompile/NetlistElaboration.cpp
```diff
@@ -1421,6 +1421,24 @@ bool NetlistElaboration::high_conn_(ModuleInstance* instance) {
       }
     }
   }
+  // Finally any unconnected ports with default value gets assigned the value
+  if (ports) {
+    for (auto p : *ports) {
+      if (p->High_conn() == nullptr) {
```
hzeller, line 1427: You can simplify the super-nestedness here by just doing an early-out

```c++
for (auto p : *ports) {
  if (p->High_count() != nullptr) continue;
    ...
  if (found == allSignalss.end()) continue;
    ...
  if (NodeId defaultId != pair.second->getDefaultValue()) continue;
  ...
}
```

... and thus, from a very complicated, high-complexity code, it is very quick to parse for the human.

src/DesignCompile/NetlistElaboration.cpp
```diff
@@ -1421,6 +1421,24 @@ bool NetlistElaboration::high_conn_(ModuleInstance* instance) {
...
+        auto itr = allSignals.find(p->VpiName());
```
hzeller, line 1428: Calling iterators that come back from a `find()` operation are much easier to understand if they are named `found` (which is what the represent) not `itr` (which is technically what the type is, but has absolutely no meaning in the context).

```c++
  auto found = allSignals.find(p->VpiName());
```

src/DesignCompile/NetlistElaboration.cpp
```diff
@@ -1421,6 +1421,24 @@ bool NetlistElaboration::high_conn_(ModuleInstance* instance) {
...
+        if (itr != allSignals.end()) {
+          auto pair = (*itr);
```
hzeller, line 1430: Do you want a copy or a reference here ?

PR #3633, opened 2023-05-06
$signed and $unsigned

alaindargelas:


QuantamHD:
@grotival this one might solve some of our issues

alaindargelas:
@QuantamHD please don't hesitate to send a bulk of mini-testcases, I'll check that we do the correct thing.
All is needed is the LHS of the expression with it's declared type and the RHS expression with a definition of the constants.
If the expression is non-constant, then there is no expression reduction in Surelog and the $signed/$unsigned is passed as a system function call to the Yosys plugin.

PR #3757, opened 2023-07-16
Enable GC by default

alaindargelas:
Trying to find what is wrong with GC


alaindargelas:
@hs-apotell , I enabled the GC by default, I can't find the issues you reported.

PR #3768, opened 2023-07-22
module def lookup hier_path

alaindargelas:


alaindargelas:
@kamilrakoczy @mglb this PR solves OpenTitan long standing hier_path resolution errors.

PR #3796, opened 2023-08-08
Make it possible to choose using host-provided capnproto

hzeller:
Now that we use capnproto in Surelog directly, provide the option (previously, we were using whatever the UHDM submodule leaked into our cmakefile).


timkpaine:
This is already lagging https://github.com/chipsalliance/Surelog/blob/master/.github/workflows/non_vendored.yml

PR #3832, opened 2023-09-02
WIP - fix vendored workflows

timkpaine:
PR opened for visibility, will close as I iterate if necessary to avoid spamming everyone.

- fixes https://github.com/chipsalliance/Surelog/issues/3831
- adds some notes to avoid naively adding a dependency on a submodule.
- pins UHDM version for tests, adds version to submodule listing so that one can easily just find/replace all instances of the UHDM version during upgrades.


alaindargelas:
https://github.com/chipsalliance/Surelog/actions/runs/6061196525 has been failing on branches, it passes on main.
Can't merge to main with this test failing.

PR #3835, opened 2023-09-03
filter sysfunc error from initial stmt in synth mode

alaindargelas:


Thomasb81:
This PR is not clear for me or may be incomplete ?

I would expect a synthesis tool to ignore system task of the kind `$fatal`, `$error`, `$warning`, `$info` whenever the kind of process they are located `initial`, `final`, `always`, `always_comb`, `always_ff`. Not only in `initial` process.

Those system task could be used as part of assertion embed in design code that is suppose to pass synthesis without code modification.

alaindargelas:
@Thomasb81 , yes, I'm not sure the extend, it looks like Yosys supports quite a few, but it does not support $error

alaindargelas:
If you have the bandwidth, please try all the permutations with the Yosys plugin and Yosys

PR #3836, opened 2023-09-03
Revert "fix clang format complaints"

alaindargelas:
Reverts chipsalliance/Surelog#3830 attempting a revert to disagnose failure of yosys plugin test


alaindargelas:
That's not it.

PR #3838, opened 2023-09-04
Revert "Issue#3812#3818 Introducing json, and cache working directories for sepcmd"

alaindargelas:
Reverts chipsalliance/Surelog#3827


alaindargelas:
This PR is not the culprit either

PR #3841, opened 2023-09-05
filter non-synth funcs

alaindargelas:


alaindargelas:
Looks like github actions are down

PR #3842, opened 2023-09-05
Fix SV plugin CI

kbieganski:
~~The SystemVerilog plugin is undergoing a big rework that aims to ease development. Part of it is improving the CI, and exposing the test workflow as a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows). Thanks to this, the Surelog CI can just call it (and override the Surelog branch in it). This means the plugin tests can run in this repo with mostly no ingerence.~~

Fixes the plugin CI.

### DIFF
--- a/.github/workflows/yosys-plugin.yml
+++ b/.github/workflows/yosys-plugin.yml
@@ -35,22 +35,26 @@ jobs:
         update-alternatives --install /usr/bin/pip pip /usr/bin/pip3 1
         pip install orderedmultidict
 
-    - name: Setup integration repository
+    - name: Setup plugin repository
       run: |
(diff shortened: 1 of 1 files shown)


kbieganski:
Unfortunately, the approach with reusable workflows is not viable. In short, it makes the CI over at [chipsalliance/systemverilog-plugin](https://github.com/chipsalliance/systemverilog-plugin) an unmaintainable mess. So I'll just fix the CI here and we'll be careful no to break it (build process improvements are incoming).

kbieganski:
The CI failures are unrelated to this change.

timkpaine:
thanks @kbieganski !

PR #3891, opened 2023-09-28
Fix submodule checkout in plugin CI

kbieganski:
https://github.com/chipsalliance/synlig/pull/2012 broke it.


kbieganski:
The CI failure is unrelated to this patch.

PR #3906, opened 2023-10-15
Bump Version: 1.76

hzeller:


alaindargelas:
@hzeller , aren't we supposed to bump here too?
https://github.com/chipsalliance/Surelog/blob/master/.gitmodules

The .gitmodules is stuck at 1.74
And I think that is the root cause for:
https://github.com/chipsalliance/Surelog/issues/3855

which is still not resolved.

hzeller:
Oh, I didn't know about the manual update of .gitmodules, I just updated the third_party/ part. Will do.

Issue #3855, opened 2023-09-13
fatal: Needed a single revision, Unable to find current origin/v1.74 revision in submodule path 'third_party/UHDM'


alaindargelas:
systemverilog-plugin/third_party/surelog$ git submodule update --remote --merge
fatal: Needed a single revision
Unable to find current origin/v1.74 revision in submodule path 'third_party/UHDM'

alaindargelas:
@timkpaine any idea?

timkpaine:
I’m trying to repro locally but theres a lot going on in “systemverilog-plugin” which is now “synlig” (??) and has also changed default branch from master to main.

alaindargelas:
Actually I got the same error doing a recursive update in Surelog

PR #3911, opened 2023-10-18
Use same version in .gitmodules as in submodules for UHDM.

hzeller:
Context: #3906 #3855


hzeller:
If it is needed to update .gitmodules on version bump, I should probably update the comment here (which is what I usually look at as a reminder how to version-bump): https://github.com/chipsalliance/Surelog/issues/3160#issuecomment-1483416441

hzeller:
I've updated the comment in https://github.com/chipsalliance/Surelog/issues/3160#issuecomment-1483416441 - does that look about right @timkpaine ?

PR #3966, opened 2024-05-07
Fix issue with linking dependency of -lstdc++fs when using flatpak

sebinho:
When trying to build Surelog inside a flatpak, it fails due to a linking issue with `-lstdc++fs`.
See opened topic here: https://github.com/flathub/flathub/issues/5221

It seems to be an old legacy linking practice that should be removed or it should be able to properly detect old versions of Ubuntu that requires it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -655,10 +655,6 @@ if (UNIX)
 endif()
 
 if (CMAKE_SYSTEM_NAME MATCHES "Linux")
-  if(NOT SURELOG_USE_HOST_UHDM)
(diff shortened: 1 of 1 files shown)


timkpaine (approved):
lgtm, agreed it seems stale and I don't think its needed

hzeller:
LGTM. Context: this was indeed a legacy leftover from the time around 2019 when compilers did not universally have the latest `std::filesystem` stuff (which should've been there in c++17), so shipped it as separate libraries.

Thanks for removing.

PR #4004, opened 2024-11-08
Fix some missing std:: includes in src/Design

hzeller:
After looking at the output of clang-tidy [misc-include-cleaner] results...

Also, whlie at it, move the foo.h header included in foo.cpp to the very top, as this is best practice to catch if the header is self-contained.


hzeller:
For now, this is only in src/Design, but I plan to do more cleanups after this first round (but then submit it myself to not overly strain the reviewers attention)

PR #4028, opened 2025-03-19
Update CI to run more recent Macos

hzeller:


hzeller:
Ubuntu is on 24.04 on all Linux but one, the Coverage one. Is this you mean ? Also, why 22 not 24 there ?

hzeller:
FYI @alaindargelas moving CI back towards working

hzeller:
Looks like the coverage stopped working in 2022 so I might as well remove that https://app.codecov.io/gh/chipsalliance/Surelog?search=&trend=all%20time

hzeller:
> > Looks like the coverage stopped working in 2022 so I might as well remove that https://app.codecov.io/gh/chipsalliance/Surelog?search=&trend=all%20time
> 
> I would recommend disabling it rather than deleting it.

OTOH, it's always simple to re-create (I added the original coverage CI job), but I prefer not to have disabled configuration that is just sitting there even more bitrotting (I mean, it has been not working for 2.5 years without anyone noticing)

PR #4029, opened 2025-03-20
Update antlr

hzeller:


hzeller:
there is something wrong with the upstream cmake but my cmake-fu is low

alaindargelas:
Worked on in https://github.com/chipsalliance/Surelog/pull/4034

PR #4034, opened 2025-05-30
Update antlr

alaindargelas:

PR #4035, opened 2025-05-30
Bump UHDM and install python package

dpetrisko:
This PR bumps UHDM to current master and adds the "UHDM_WITH_PYTHON" flag to the Surelog python build. This means that users can immediately access the UHDM python API as so:

```
make release_with_python
...
python -c "from uhdm import uhdm"
```


dpetrisko:
Looks like the failing windows build error is in antlr:

```
D:\a\Surelog\Surelog\third_party\antlr4\runtime\Cpp\runtime\src\atn\ProfilingATNSimulator.cpp(38): error C2653: 'high_resolution_clock': is not a class or namespace name
```

which appears to be the same as #4034 . I assume once that PR is fixed, can rebase on top and be good to go!

PR #4046, opened 2025-07-11
fix: don't try to attribute names to ordered ports

Rodrigodd:
The code was attributing name to ordered ports based on the Primary literal passed to the port of the module instantiation (if any), which is not the name of the port. I believe the port name is later corrected when the module is resolve to its declaration. But if the module is not resolve (is a unknown module) the generated UHDM would have the wrong name for the ports. It is better to keep the port unnamed.

Fix https://github.com/chipsalliance/synlig/issues/2767.


alaindargelas:
@Rodrigodd , thanks for the PR!

PR #4058, opened 2025-07-30
Fix a bunch of IWYU issues

hzeller:
... reducing `misc-include-cleaner` warnings (observed when running `.github/bin/run-clang-tidy-cached.cc`) from 340-ish to under 10.


hzeller:
Don't know what is up with the windows regression, but that looks like a pre-existing issue (some sort of double-newline became one newline ?)

PR #4092, opened 2026-06-09
for regression, allow the path to the surelog folder to contain dot

fabrice-baray:
Hi,

I was unable to run the regressions because the folder in which surelog was cloned was containing a dot in its name. I suggest by that change to allow the folder to have a dot.

Btw I believe that in tcl regexp, in the square brackets, we do not need to escape the dot. I kept it escaped because it was in the other part of the regexp, to stay consistent. But both should work.

Regards,
  Fabrice

### DIFF
--- a/tests/regression.tcl
+++ b/tests/regression.tcl
@@ -315,12 +315,12 @@ proc load_tests { } {
     set totaltest 0
     foreach file $fileList {
         if {$VERIFICATION == 0} {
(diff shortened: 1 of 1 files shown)


alaindargelas:
@fabrice-baray thanks, I'll merge it, but know that the regression has now been changed to use:
python3 scripts/regression.py run 
we don't use the Tcl version anymore

fabrice-baray:
Thanks! I didn't know about python version. Very good I will use that one now.

PR #4099, opened 2026-07-05
PreprocessFile: bound the token-paste look-ahead to avoid OOB read

EylonKrause:
### Problem
`PreprocessFile` reads a token vector out of bounds when the `` ` `` `` ` `` token-paste operator is the last body token.

### Root cause
In `SourceCompile/PreprocessFile.cpp`, the ``s1 == "``"`` branch guards the previous-token access (`if (i > 0) s1 = tokens[i - 1];`) but reads the next token as `tokens[i + 1]` unconditionally. When the paste operator is the last token, `i + 1 == tokens.size()`, an out-of-bounds `std::vector::operator[]`.

### Fix
Guard `tokens[i + 1]` symmetrically with the `i - 1` case.

### Verification
Static reasoning. (Surelog builds with cmake and large vendored submodules; not built locally.)


EylonKrause:

PR #4100, opened 2026-07-05
StringUtils: index the separator table with an unsigned byte

EylonKrause:
### Problem
`StringUtils::tokenize` / `tokenizeBalanced` index a `std::array<bool, 256>` with a sign-extended `char`.

### Root cause
`separators[(int32_t)ch]` (and `separators[(int32_t)str[i]]`): a plain `char` is signed on most platforms, so an input byte ≥ `0x80` (non-ASCII / UTF-8 / binary source text) sign-extends to a negative `int32_t` and indexes the array out of bounds.

### Fix
Cast to `unsigned char` at all four sites.

### Verification
Static reasoning; textbook signed-`char` sign-extension. (Not built locally.)

### DIFF
--- a/src/Utils/StringUtils.cpp
+++ b/src/Utils/StringUtils.cpp
@@ -93,14 +93,14 @@ std::vector<std::string_view>& StringUtils::tokenize(
   if (str.empty()) return result;
 
   std::array<bool, 256> separators = {false};
-  for (char ch : any_of_separators) {
-    separators[(int32_t)ch] = true;
+  for (unsigned char ch : any_of_separators) {
+    separators[ch] = true;
   }
 
   size_t start = 0;
   size_t end = 0;
-  for (char ch : str) {
-    if (separators[(int32_t)ch]) {
+  for (unsigned char ch : str) {
+    if (separators[ch]) {
       result.emplace_back(str.data() + start, end - start);
       end = start = end + 1;
     } else {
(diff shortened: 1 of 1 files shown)


src/Utils/StringUtils.cpp
```diff
@@ -94,13 +94,13 @@ std::vector<std::string_view>& StringUtils::tokenize(
 
   std::array<bool, 256> separators = {false};
   for (char ch : any_of_separators) {
```
hzeller: would it be possible to simply say `for (unsigned char : ...)` here ?

src/Utils/StringUtils.cpp
```diff
@@ -94,13 +94,13 @@ std::vector<std::string_view>& StringUtils::tokenize(
...
-    separators[(int32_t)ch] = true;
+    separators[(unsigned char)ch] = true;
```
hzeller: ... if not, I'd probably cast to `uint8_t` as it improves readability

EylonKrause:
Thanks @hzeller! Switched the two build loops to `for (unsigned char ch : ...)` so the separator table is indexed directly with no cast, and used a `uint8_t` cast for the one index-based access (`str[i]` in `tokenizeBalanced`, where a loop variable isn't available). Reads cleaner now.

EylonKrause:

PR #4101, opened 2026-07-06
CommandLine: parse numeric arguments without throwing

EylonKrause:
### Problem
Several numeric command-line arguments are parsed with an unguarded `std::stoi`, which throws (`std::invalid_argument` / `std::out_of_range`) on non-numeric or oversized input. There is no `try`/`catch` on the command-line path (`main` → `parseCommandLine`), so a trivial typo terminates Surelog with an uncaught C++ exception instead of a diagnostic. A separate unsigned-underflow also causes an out-of-bounds read in `-cmd_ign`.

Reproducers (all currently crash):
- `surelog -mt auto file.sv` / `surelog -mt 99999999999 file.sv`
- `surelog -split foo file.sv`
- `surelog -d .` / `surelog -d 99999999999999999999`
- `surelog -cmd_ign cmd notanumber` (throw), and `surelog -cmd_ign` alone (see below)

### Details
- `-mt`/`-mp`/`--threads` (`maxMT = std::stoi(...)`) and `-split` (`m_nbLinesForFileSplitting = std::stoi(...)`) have no numeric guard.
- `-d` gates on `is_number()`, but `is_number` is `find_first_not_of("-.0123456789") == npos`, which accepts `"."`, `"-"`, and `> INT_MAX` digit strings, so `std::stoi` still throws.
- `-cmd_ign` (`processArgs_`): `if (i < args.size() - 2)` computes an **unsigned** subtraction — for a lone `-cmd_ign` token (`args.size() == 1`), `1 - 2` wraps to `SIZE_MAX`, the guard passes, and `args[i + 1]`/`args[i + 2]` read a 1-element vector out of bounds. The count is then `std::stoi(args[i + 2])`, which also throws on a non-numeric value.

### Fix
Route all four numeric parses through the codebase's own non-throwing `NumUtils::parseInt32`/`parseUint32` (from_chars-based; returns `nullptr` on failure), emitting the handler's existing diagnostic (`CMD_MT_INCORRECT_LEVEL`, `CMD_SPLIT_FILE_MISSING_SIZE`, `CMD_DEBUG_INCORRECT_LEVEL`, or the `-cmd_ign` cerr message) on failure. Also fix the `-cmd_ign` guard to `i + 2 < args.size()` to avoid the unsigned underflow.

### Test
`InvalidNumericArgumentsDoNotCrash` in `CommandLineParser_test.cpp` runs `parseCommandLine` over each of the above argument sets and requires it to return without throwing (each previously terminated the process).

### DIFF
--- a/src/CommandLine/CommandLineParser.cpp
+++ b/src/CommandLine/CommandLineParser.cpp
@@ -49,6 +49,7 @@
 #include "Surelog/ErrorReporting/ErrorDefinition.h"
 #include "Surelog/ErrorReporting/Location.h"
 #include "Surelog/SourceCompile/SymbolTable.h"
+#include "Surelog/Utils/NumUtils.h"
 #include "Surelog/Utils/StringUtils.h"
 #include "Surelog/surelog-version.h"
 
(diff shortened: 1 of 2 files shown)


EylonKrause:

PR #4102, opened 2026-07-07
SourceCompile: don't throw on an oversized `timescale value

EylonKrause:
### Problem
`SV3_1aTreeShapeListener::enterTimescale_directive` parses the `` `timescale `` unit/precision values with `std::stoi`, which throws `std::out_of_range` on a value exceeding `int` range — crashing the parser with an uncaught exception.

### Details
The grammar's `Integral_number` allows an unbounded digit string, and the regex group `([0-9]+)` captures it verbatim, so a directive like `` `timescale 3000000000 s / 1 s `` (any size/precision `> INT_MAX`) makes `std::stoi(base1)` / `std::stoi(base2)` throw. The parse-tree walk (`ParseTreeWalker::walk`, `ParseFile.cpp:529/568`) is not wrapped in a `try`/`catch` on this path — the only catch blocks handle `antlr4::ParseCancellationException` around the parse itself — so the exception unwinds to `std::terminate`.

The preprocessor-side sibling (`SV3_1aPpTreeShapeListener::enterTimescale_directive`, lines 1041/1045) already parses the same values with non-throwing `atoi`.

### Fix
Use non-throwing `atoi` for both the time-unit and precision values, matching the Pp-side sibling. An overflowed value is still `!= 1/10/100`, so it is reported through the existing `PA_TIMESCALE_INVALID_VALUE` diagnostic instead of crashing.

### Testing
Verified by static reasoning (`std::stoi` overflow semantics). Happy to add an end-to-end parse test for an oversized `` `timescale `` value if you'd like — kept this to the minimal sibling-matching fix.

### DIFF
--- a/src/SourceCompile/SV3_1aTreeShapeListener.cpp
+++ b/src/SourceCompile/SV3_1aTreeShapeListener.cpp
@@ -28,6 +28,7 @@
 
 #include <cctype>
 #include <cstdint>
+#include <cstdlib>
 #include <regex>
 #include <string>
 #include <string_view>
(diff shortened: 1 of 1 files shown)


EylonKrause:

